PR #5, opened 2020-02-24
vmware_rest_client is in the vmware collection

goneri:
plugins/module_utils/vmware_rest_client.py was imported by error. It
should be in the `community/vmware` collection.


goneri:
recheck

PR #6, opened 2020-02-25
move VmwareRestModule two level below

goneri:
Move `ansible_collections.ansible.vmware_rest.plugins.module_utils.vmware_httpapi.VmwareRestModule.VmwareRestModule` in
`ansible_collections.ansible.vmware_rest.plugins.module_utils.vmware_httpapi.VmwareRestModule` to reduce the path
length.


goneri:
re

PR #38, opened 2020-08-20
depends on cloud.common

goneri:
Use the cloud.common collection to share what was previously
in `ansible_module.turbo`.

Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/628


goneri:
recheck

goneri:
recheck

PR #59, opened 2020-08-28
improve idempotency and changed flag

goneri:
- set an `id` key when possible
- properly set `changed` flag for some modules. e.g: `vcenter_vm_hardware_boot

version: 2d80821818c23350d48773ad34a1da981d51a883


goneri:
recheck

PR #60, opened 2020-08-28
add abilty to modify existing resource

goneri:
- Add ability to modify existing devices.
- Use `absent` and `present` states instead of the operation names
  (`create`, `update`, `delete`).

version: 23ddadb1a3e642c7c41a91962597e309262c42c6


goneri:
recheck

goneri:
recheck

PR #66, opened 2020-09-01
drop tech previous modules

goneri:
Remove two modules because they depend on end-points with the Tech
Preview status.


goneri:
recheck

PR #70, opened 2020-09-02
tests: extend the functional tests

goneri:
Depends-On: https://github.com/ansible-collections/vmware_rest/pull/73


goneri:
recheck

PR #127, opened 2020-10-01
tests: refactoring of the test directory

goneri:
- split up the vcenter_vm_scenario1 role and create a new role called
  prepare_lab
- add a new play dedicated to collect the lab information


goneri:
recheck

Issue #153, opened 2020-10-20
Cannot create a resource pool following current documentation


matthewcosgrove:
##### SUMMARY
Cannot create a resource pool following current documentation

From what I gather, I need to provide the name of the new resource pool and it's parent. This is an accordance with all documented parameters which state `Required with state=['present']`

My choice of parent comes from viewing the vcenter UI via Hosts and Clusters. Within the playbook below I retrieve the moid of the cluster I want to create the resource pool under (the cluster name was already set where I am running the playbook via the env var GOVC_CLUSTER so I just reuse that to filter on)

So far I have 

```
- name: Create resource pool
  hosts: localhost
  vars:
    existing_cluster: "{{ lookup('env', 'GOVC_CLUSTER') }}"

  collections:
  - vmware.vmware_rest
  tasks:
  - name: "get parent cluster for resource pool"
    vcenter_cluster_info:
      filter_names:
      - "{{ existing_cluster }}"
    register: my_cluster
  - name: "ensure resource pool created"
    vcenter_resourcepool:
      name: "new-resource-pool"
      parent: "{{ my_cluster.value[0].cluster }}"
    register: my_resource_pool
  - debug: msg="{{ my_resource_pool }}"
```

but this results in 

```
fatal: [localhost]: FAILED! => {
    "_debug_info": {
        "operation": "create",
        "status": 404
    },
    "changed": false,
    "invocation": {
        "module_args": {
            "cpu_allocation": null,
            "memory_allocation": null,
            "name": "new-resource-pool",
            "parent": "domain-c2050",
            "resource_pool": null,
            "state": "present",
            "vcenter_hostname": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "vcenter_password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "vcenter_rest_log_file": null,
            "vcenter_username": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "vcenter_validate_certs": false
        }
    },
    "type": "com.vmware.vapi.std.errors.not_found",
    "value": {
        "error_type": "NOT_FOUND",
        "messages": [
            {
                "args": [],
                "default_message": "Resource pool with identifier '{}' does not exist.",
                "id": "com.vmware.api.vcenter.resourcepool.not_found"
            }
        ]
    }
}
```

[...]

PR #155, opened 2020-10-20
test: add functional test coverage for resource pool

goneri:
Add documentation for resource pool.

Closes: #153

### DIFF
--- a/changelogs/fragments/resourcepool.yaml
+++ b/changelogs/fragments/resourcepool.yaml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+- vcenter_resourcepool - add example in documentation.
+- vcenter_resourcepool_info - add example in documentation.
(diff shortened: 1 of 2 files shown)


matthewcosgrove:
I provided some observations related to this example over in issue #153 but probably better to add the suggestion directly here as I think it got lost there in the bigger picture.

This new example shows how to create an ad-hoc resource pool, but _not_ how to create a resource pool under a specific cluster. It looks like the intention was there to do the latter because you have the task "get parent cluster for resource pool" but then never use it to demonstrate the exact moid required for the parent.

I shared an example which does grab the required parent moid to associate it with a specific cluster if its useful on https://github.com/ansible-collections/vmware.vmware_rest/issues/153#issuecomment-713461758

Which includes the key lines
```
  - name: "ensure resource pool created"
    vcenter_resourcepool:
      name: "new-resource-pool"
      parent: "{{ my_cluster_details.value.resource_pool }}"
```

PR #184, opened 2021-04-23
Add bindep.txt for execution environments

pabelanger:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/185

Because we need to build aiohttp, some dependencies require gcc to
compile those python packages.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>


goneri:
recheck

PR #191, opened 2021-05-13
tests: re-enable appliance_services tests

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/926


goneri:
recheck

goneri:
recheck

PR #193, opened 2021-05-14
tests: one single health play

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/192

Move the `appliance_health_applmgmt_info` in the main heath play.


goneri:
recheck

PR #194, opened 2021-05-14
shutdown VM two times

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/192

If we shutdown a VM two times in a row, the second call raises an
`ALREADY_IN_DESIRED_STATE` that we can ignore.


goneri:
recheck

PR #197, opened 2021-05-17
manual: use sphinx-ansible

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/200

- switch to sphinx_ansible_theme
- enable sphinx-ansible
- fix a couple of errors
- expose the environment var to the tox environment
- drop the task_outputs files!


goneri:
recheck

PR #204, opened 2021-05-18
init.sh and init.fish files

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/932

- add init.fish
- ensure the init.sh script works with absolute paths.


goneri:
recheck

goneri:
recheck

PR #220, opened 2021-06-08
content_library: simplify the tests

goneri:
- Reduce the amount of YAML
- Drop some debug tasks
- Avoid some transitional tasks


Akasurde (approved):
LGTM

PR #228, opened 2021-06-23
bump auto-generated modules

goneri:
version: 2.0.0


pabelanger:
@goneri so for this PR, I believe you'll need to close and create a new one. Github really doesn't provide a way to clear checks from the PR level.

This is due to change in zuul 4.5.1, which has been fixed but there are 1 or 2 PRs that before we could land the fix.

goneri:
ok, thanks.

goneri:
https://github.com/ansible-collections/vmware.vmware_rest/pull/229

PR #230, opened 2021-07-19
Update README

Akasurde:
##### SUMMARY

Update README for Ansible inclusion

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ A collection may contain metadata that identifies these versions.
 PEP440 is the schema used to describe the versions of Ansible.
 <!--end requires_ansible-->
 
+## Python version compatibility
+
+As the `aiohttp` for Python has [requirement](https://github.com/aio-libs/aiohttp#requirements) of Python 3.7 or greater, this collection requires Python 3.7 or greater.
+
 ## Installation and Usage
 
 ### Install the dependencies
@@ -185,7 +189,6 @@ Name | Description
 
 <!--end collection content-->
 
-
 ### Documentation
 
 The [VMware REST modules guide](https://docs.ansible.com/ansible/devel/scenario_guides/guide_vmware_rest.html) gives a step by step introduction of the collection.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,12 +212,39 @@ ansible-test network-integration --python 3.7 --inventory /tmp/inventory-vmware_
...
+
 ## Communication
 
 We have a dedicated Working Group for VMware.
 You can find other people interested in this in `#ansible-vmware` on [libera.chat](https://libera.chat/) IRC.
 For more information about communities, meetings and agendas see https://github.com/ansible/community/wiki/VMware.
 
+## Code of Conduct
+
+This repository adheres to the [Ansible Community code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html)
```
tadeboro, line 237: This is already present in the readme file.

tadeboro (changes requested):

README.md
```diff
@@ -14,6 +14,10 @@ A collection may contain metadata that identifies these versions.
 PEP440 is the schema used to describe the versions of Ansible.
 <!--end requires_ansible-->
 
+## Python version compatibility
+
+As the `aiohttp` for Python has [requirement](https://github.com/aio-libs/aiohttp#requirements) of Python 3.7 or greater, this collection requires Python 3.7 or greater.
```
goneri, line 19: Actually, it's py3.6, not 3.7.

Akasurde: aiohttp requires 3.7 and onwards.

tadeboro: The stable version requires Python 3.6, development 3.7. But none of this answers the question from my review: why was Python 3 selected as the minimum version while most other community collections still need to maintain compatibility with 2.7 if at all possible.

jillr: @tadeboro Due to the requirements of the dependent library (aiohttp).  Most of the cloud-related collections are 3-only due to dependencies, my understanding is that documenting this should be adequate for inclusion.

jillr: @Akasurde in the community meeting we discussed this and the confusion is why we are requiring aiohttp over other http connection methods available in core anisble.  Can you please update this PR to include a note to the effect that we require python3, because we require aiohttp, because it adds major performance improvements?

tadeboro: Were collection requirements known at the time when the collection was initially designed? If yes, then aiohttp library could possibly be avoided and the collection made Python 2.7 compatible. Otherwise, the dependency argument holds and things are all good.

jillr: The 1.0.0 release of the collection was Oct 2020 and development started no later than Feb 2020 (I'm pretty sure it was even before then though).  We started working on inclusion requirements around Aug 2020 I think?   But didn't really get the Python requirements until March 2021.

tadeboro: Thanks @jillr! I will ad this to info to my review and stop nagging you all about Python versions ;)

goneri (approved):

PR #233, opened 2021-07-27
tests: skip appliance_services_info

goneri:
appliance_services_info with no parameters fails because of a problem at the vCenter level.


goneri:
Superseded by https://github.com/ansible-collections/vmware.vmware_rest/pull/234

PR #263, opened 2021-10-05
tests: use the local (ESXi) datastore

goneri:
Improve the reliability of the test by using the datastore of the ESXi.


goneri:
recheck

goneri:
recheck

Issue #267, opened 2021-10-21
vcenter_vm_hardware_ethernet update fails for version 2.0.0 and 2.1.0


mirceahuss:
I'm using vcenter_vm_hardware_ethernet to move a vm to a different network, which works for version 1.0.2-dev16. 
However it fails for version 2.0.0 and 2.1.0.

- name: move the VM on a different network
  vmware.vmware_rest.vcenter_vm_hardware_ethernet:
    vm: vm-8173
    nic: 4000
    backing:
      type: STANDARD_PORTGROUP
      network: network-21

I'm running ansible 2.9.23 on Centos 8.4
Vcenter 7.0.3

Running the task above on version 2.0 and higher results in the following error message:
fatal: [vm_name]: FAILED! => {"changed": false, "id": "4000", "value": {"error_type": "INVALID_ARGUMENT", "messages": [{"args": ["com.vmware.vcenter.vm.hardware.ethernet.update"], "default_message": "Unable to validate input to method com.vmware.vcenter.vm.hardware.ethernet.update", "id": "vapi.invoke.input.invalid"}, {"args": ["spec"], "default_message": "Found invalid field spec in structure.", "id": "vapi.data.structure.field.invalid"}, {"args": ["backing"], "default_message": "Found invalid field backing in structure.", "id": "vapi.data.structure.field.invalid"}, {"args": [], "default_message": "The given OptionalValue does not match the OptionalDefinition", "id": "vapi.data.optional.validate"}, {"args": ["com.vmware.vcenter.vm.hardware.ethernet.backing_spec", "type"], "default_message": "Structure 'com.vmware.vcenter.vm.hardware.ethernet.backing_spec' is missing a field: type", "id": "vapi.data.structure.field.missing"}]}}

I'm guessing from the error message that ethernet.backing thinks it's missing a field ?

PR #272, opened 2021-10-29
bump auto-generated modules

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/271
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1207
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/274
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1210

The modules are autogenerated by:
https://github.com/ansible-collections/vmware_rest_code_generator
version: fbfc659e75b952ca5f69a1b9ff11b4db8f761675

PR #273, opened 2021-10-29
ensure subkeys integrity is preserved

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/272

Closes: https://github.com/ansible-collections/vmware.vmware_rest/issues/267


goneri:
recheck

PR #274, opened 2021-11-08
prepare_lab: wait till the library backend is ready

goneri:
With vSphere 7.0.3, we've got to wait a bit for the library backend.


goneri:
recheck

PR #282, opened 2021-11-29
lookup: use the FQCN in examples

goneri:
Use the FQCN of the built-in modules.

See: https://github.com/ansible-collections/ansible-inclusion/discussions/24


goneri:
recheck

PR #283, opened 2021-11-29
tests: use FQCN in the tests

goneri:
Ensure we use the full FQCN during the tests, this way we will also
get the FQCN inthe EXAMPLES blocks of the modules.


goneri:
recheck

PR #285, opened 2021-11-30
tests: use include_tasks

goneri:
The log output of the `include_tasks` calls give a better understand on
what's going


goneri:
recheck

goneri:
recheck

goneri:
recheck

PR #305, opened 2022-01-18
adjust the version_added field of the lookup plugins

goneri:
See: https://github.com/ansible-collections/ansible-inclusion/discussions/24#discussioncomment-1977703


changelogs/fragments/version_added.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- "Adjust the release version of the lookup plugins fro, 2.0.1 to 2.1.0."
```
felixfontein, line 3:
```suggestion
- "Adjust the release version of the lookup plugins from 2.0.1 to 2.1.0." 
```

PR #314, opened 2022-04-07
Upgrade and pin black version

Akasurde:
##### SUMMARY

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tox.ini

PR #313, opened 2022-04-07
Sanity tests: copy ignore-2.13.txt to ignore-2.14.txt

Andersson007:
##### SUMMARY
https://github.com/ansible-collections/news-for-maintainers/issues/13

Dear maintainers,

I'd like to ask you to track the news-for-maintainers themselves in the future if possible

Thanks!

PR #317, opened 2022-04-22
Resolve black and sanity errors

jillr:
##### SUMMARY
Add sanity ignore for ansible 2.14 and upgrade black version

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tox.ini
tests/ignores/sanity-2.14.txt

This would supersede #313 and #314 

##### ADDITIONAL INFO

Versions of `black` newer than 19.x introduce formatting changes that will need to be incorporated into the generated modules before we can check against those versions in CI
https://black.readthedocs.io/en/stable/change_log.html#id61
Changes in `click` require `black>=22.3.0`.
https://github.com/psf/black/issues/2964

For now let's pin both at what works.


jillr:
recheck

jillr:
build failures are hitting disk out of space issues.  I don't have access to vexxhost so rechecking until I get a good node.

jillr:
recheck

PR #326, opened 2022-05-03
integration: share the same disk image between targets

goneri:
Avoid the need to redownload the RHEL image in every test. Right
now the image is only used in the vcenter_vm_customize test.


mariolenz:
regate

PR #335, opened 2022-06-10
EXPERIMENTAL: adjustment for an asyncio based executor

goneri:


mariolenz:
I think we can safely close this two year old PR.

PR #342, opened 2022-06-17
appliance/dns: adjustment for VCSA 7.0U3c-19193900

goneri:
VCSA 7.0U3c-19193900 behavior has changed a bit and we cannot remove DNS
server set by DHCP anymore.


goneri:
regate

mariolenz:
regate

PR #345, opened 2022-06-22
test/prepare_lab: wait just up to 15s

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1576
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1577

300s was too long.


goneri:
recheck

Issue #364, opened 2022-11-15
Multiple vCenters at the same time can cause errors


MallocArray:
##### SUMMARY

Might be a repeat of #336 where playbooks with multiple tasks that use the vmware.vmware_rest modules against multiple vCenters at the same time may error.  It is not completely repeatable but does happen often.  Sometimes a retry fixes it.  Unable to reproduce with a single vCenter.  Running cloud.common 2.1.2 and vmware.vmware_rest 2.2.0

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Observed in
appliance_access_ssh
appliance_networking_proxy
and possibly others
##### ANSIBLE VERSION

```paste below
ansible [core 2.13.6]
  config file = /runner/ansible.cfg
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  ansible collection location = /runner/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.13 (default, Jun 24 2022, 15:27:57) [GCC 8.5.0 20210514 (Red Hat 8.5.0-13)]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection         Version
------------------ -------
vmware.vmware_rest 2.2.0 
cloud.common 2.1.2  
```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/runner/ansible.cfg) = ['/runner/collections']
DEFAULT_FILTER_PLUGIN_PATH(/runner/ansible.cfg) = ['/runner/custom_filters']
DEFAULT_ROLES_PATH(/runner/ansible.cfg) = ['/runner/roles']
DEFAULT_STRATEGY_PLUGIN_PATH(/runner/ansible.cfg) = ['/runner/custom_plugins/mitogen-0.3.0-rc.0/ansible_mitogen/plugins/strategy']
DEFAULT_TIMEOUT(/runner/ansible.cfg) = 40
HOST_KEY_CHECKING(/runner/ansible.cfg) = False
PARAMIKO_LOOK_FOR_KEYS(/runner/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Ansible Execution Environment from ansible-builder
vCenter 7.0 U3

##### STEPS TO REPRODUCE

Playbook with several tasks in a row all using the same collection.  After 

```yaml
[...]

PR #373, opened 2022-12-01
appliance-multi-hosts tests

goneri:
Test to validate we can run the same module in parallel.

See: https://github.com/ansible-collections/vmware.vmware_rest/issues/364


alinabuzachis:
recheck

alinabuzachis (approved):

PR #374, opened 2022-12-01
tests: rename the YAML file with the .yml extension

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/376

The goal is to reduce the delta between this collection and amazon.cloud which
is also auto-generated. By keeping the differences minimal, we can reuse the
same tools to maintain the two collections.


goneri:
regate

PR #410, opened 2023-04-20
CI - linters and changelog github actions workflow

abikouo:
CI migration from zuul to GHA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: CI
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - labeled
+      - unlabeled
+      - synchronize
+    branches:
+      - main
+      - stable-*
+    tags:
+      - '*'
+
+jobs:
+  changelog:
+    uses: ansible-network/github_actions/.github/workflows/changelog.yml@main
+  linters:
+    uses: abikouo/github_actions/.github/workflows/tox-linters.yml@tox_linters
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Linters
 concurrency:
   group: ${{ github.head_ref }}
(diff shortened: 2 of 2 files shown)


abikouo:
recheck

alinabuzachis (approved):

mariolenz:
There already is a test for a changelog: #406 / #433

For the linter test, I don't understand why you use `abikouo/github_actions/.github/workflows/tox-linters.yml@tox_linters`. It looks like there is one in [ansible-network/github_actions](https://github.com/ansible-network/github_actions). I think it would be better to use this one.

Actually, I think it would be best to have linting in [ansible-collections/collection_template](https://github.com/ansible-collections/collection_template) and we could just copy it it to this collection. Maybe you should create a PR there to implement a default linting CI workflow ;-)

abikouo:
> There already is a test for a changelog: #406 / #433
> 
> For the linter test, I don't understand why you use `abikouo/github_actions/.github/workflows/tox-linters.yml@tox_linters`. It looks like there is one in [ansible-network/github_actions](https://github.com/ansible-network/github_actions). I think it would be better to use this one.
> 
> Actually, I think it would be best to have linting in [ansible-collections/collection_template](https://github.com/ansible-collections/collection_template) and we could just copy it it to this collection. Maybe you should create a PR there to implement a default linting CI workflow ;-)

@mariolenz this PR needs some updates. 
Indeed, the changelog workflow added here is not valid anymore as there is already one.
And yes we should not reference action from personal repos.

mariolenz (approved):
I'm no GHA expert but as far as I can see this LGTM.

@alinabuzachis What do you think?

alinabuzachis (approved):

mariolenz:
> Depends-On: [ansible/ansible-zuul-jobs#1789](https://github.com/ansible/ansible-zuul-jobs/pull/1789)

I've removed this. I think it makes more sense if removing those tests from Zuul should depend on implementing them here in the collection instead of the other way around ;-)

mariolenz:
regate

mariolenz:
Cool, we now also have a GHA linting workflow. Thanks @abikouo!

PR #416, opened 2023-05-01
Module generated with changes to ansible.content_builder

GomathiselviS:
##### SUMMARY

The vmware header template in [Content Builder tool](https://github.com/ansible-community/ansible.content_builder) is updated with [imports and pylint skips](https://github.com/ansible-community/ansible.content_builder/pull/58). This PR has the changes made to vmware_rest modules generated by the updated Content Builder tool.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/appliance_access_consolecli.py
```diff
@@ -102,8 +103,8 @@
     "set": {"query": {}, "body": {"enabled": "enabled"}, "path": {}}
 }  # pylint: disable=line-too-long
 
-import json
-import socket
+import json  # pylint: disable=unused-import
```
abikouo, line 106: I would rather prefer the generator to import only required modules

abikouo (changes requested):
Is it easy to update the content builder to import only required modules?

plugins/modules/appliance_access_consolecli.py
alinabuzachis: Agree!

plugins/modules/appliance_monitoring_info.py
```diff
@@ -82,834 +82,6 @@
 """
 
 RETURN = r"""
-# content generated by the update_return_section callback# task: Get the list of the monitored items
-value:
-  description: Get the list of the monitored items
-  returned: On success
-  sample:
-  - category: com.vmware.applmgmt.mon.cat.disk
-    description: com.vmware.applmgmt.mon.descr.disk.read.rate.dm-1
```
alinabuzachis, line 91: Examples generation does not work at the moment in the content builder, so we will end up deleting examples. We should not delete/alter the examples that already exist unless they receive an update.

plugins/modules/appliance_vmon_service_info.py
```diff
@@ -85,400 +85,15 @@
 """
 
 RETURN = r"""
-# content generated by the update_return_section callback# task: Get information about a VMON service
-value:
-  description: Get information about a VMON service
```
alinabuzachis, line 90: Same here.

plugins/modules/content_locallibrary.py
```diff
@@ -369,882 +369,12 @@
 """
 
 RETURN = r"""
-# content generated by the update_return_section callback# task: Delete all the local libraries
-msg:
```
alinabuzachis, line 373: Same here.

plugins/modules/vcenter_vm_hardware_cdrom_info.py
```diff
@@ -75,7 +75,7 @@
         type: str
 author:
 - Ansible Cloud Team (@ansible-collections)
-version_added: 0.1.0
+version_added: 2.0.0
```
alinabuzachis, line 78: This should not be updated.

plugins/modules/vcenter_vm_hardware_cpu.py
```diff
@@ -100,7 +100,7 @@
         type: str
 author:
 - Ansible Cloud Team (@ansible-collections)
-version_added: 0.1.0
+version_added: 2.0.0
```
alinabuzachis, line 103: Same here.

plugins/modules/vcenter_vm_hardware_cpu.py
```diff
@@ -129,21 +129,6 @@
 """
 
 RETURN = r"""
-# content generated by the update_return_section callback# task: Dedicate one core to the VM
```
alinabuzachis, line 132: Same here.

plugins/modules/vcenter_vm_hardware_cpu_info.py
```diff
@@ -66,7 +66,7 @@
         type: str
 author:
 - Ansible Cloud Team (@ansible-collections)
-version_added: 0.1.0
+version_added: 2.0.0
```
alinabuzachis, line 69: Same here.

plugins/modules/vcenter_vm_hardware_disk.py
```diff
@@ -150,7 +150,7 @@
         type: str
 author:
 - Ansible Cloud Team (@ansible-collections)
-version_added: 0.1.0
+version_added: 2.0.0
```
alinabuzachis, line 153: Same.

plugins/modules/vcenter_vm_hardware_disk_info.py
```diff
@@ -75,7 +75,7 @@
         type: str
 author:
 - Ansible Cloud Team (@ansible-collections)
-version_added: 0.1.0
+version_added: 2.0.0
```
alinabuzachis, line 78: Same.

alinabuzachis (changes requested):
Examples generation does not work with the ansible.content_builder and we should not alter/delete the existing examples unless there's an update. We should also not update version_added. As Aubin suggested, I would import the libraries where necessary instead of skipping the lint.

plugins/modules/appliance_access_consolecli.py
GomathiselviS: We have a single header template file, which is rendered into all the module files. Some modules use the imported libraries and some do not. Instead of skipping the pylint in the individual module files, I think adding the pylint disable in the template file is less time-consuming. What are your thoughts.?

GomathiselviS:
> Examples generation does not work with the ansible.content_builder and we should not alter/delete the existing examples unless there's an update. We should also not update version_added. As Aubin suggested, I would import the libraries where necessary instead of skipping the lint.

@alinabuzachis As I have explained in the reply to @abikouo's comment, I see that these imported libraries are not common across all module files. As we use a single template file for all the module files, it is better to mention all the libraries in the template file, which will be rendered as multiple files. Let me know your thoughts.

I have fixed the examples, return block and version_added fields.

mariolenz:
I've fixed the sanity tests in #432 by removing the unused imports with [autoflake](https://pypi.org/project/autoflake/). Therefor I think this PR is unnecessary now.

PR #429, opened 2023-11-01
[WIP] Prepare 3.0.0 release

mariolenz:
This collection still supports some outdated ansible-core versions. Since I consider dropping them a breaking change, I'd like to propose a new major release. There are no changes to or new modules, but while I'm at it I

- Removed old changelog fragments
- Configured to automatically remove changelog fragments
- Use GHA for some tests
- Truncated the changelog so it starts fresh with 3.0.0

PR #430, opened 2023-11-01
Remove old changelog fragments

mariolenz:
##### SUMMARY
There are some old changelog fragments still around that I think should be removed. I also think changelog fragments should be automatically removed when a new changelog is created.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
changelogs/config.yaml

##### ADDITIONAL INFORMATION
I wanted to propose a new major release, but I ran into so many issues in #429 that I decided to split those changes into several PRs.


alinabuzachis (approved):

mariolenz:

PR #431, opened 2023-11-01
Start using GHA for sanity tests

mariolenz:
Start using GHA for sanity tests.

_edit:_
I wanted to propose a new major release, but I ran into so many issues in #429 that I decided to split those changes into several PRs.


mariolenz:
What do you think @alinabuzachis? If this OK for you, we can merge it and I would try to create a PR removing the 2.13 and 2.14 sanity tests from the Zuul job.

mariolenz:

mariolenz:
@alinabuzachis This is weird, the `Sanity (Ⓐstable-2.15)` succeeded in the PR but failed after I've merged. I'll ask around if someone has an idea why.

BTW and FYI: ansible-community/ansible.content_builder#67

PR #432, opened 2023-11-03
Fix GHA sanity tests

mariolenz:
Fix GHA sanity tests.

Follow-up to #431

_edit_:

I know those files are generated and shouldn't be touched manually, but I'd like to see the CI succeeding. Most of the stuff were unused imports. In the long run, I think we need a fix for this: ansible-community/ansible.content_builder#67

However, there were also two issues that we should have a closer look at:

```
ERROR: Found 2 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/vcenter_vm_hardware_adapter_sata.py:0:0: doc-default-does-not-match-spec: Argument 'bus' in argument_spec defines default as (0) but documentation defines default as (None)
ERROR: plugins/modules/vcenter_vm_hardware_adapter_scsi.py:0:0: doc-default-does-not-match-spec: Argument 'bus' in argument_spec defines default as (0) but documentation defines default as (None)
```

Might be a bug in the code generator.

_edit2_:

I'll remove the 2.16 sanity tests again because:

```
Running sanity test "import" on Python 3.12
Run command with data: importer.py
Run command with data: importer.py
ERROR: Found 132 import issue(s) on python 3.12 which need to be resolved:
ERROR: plugins/modules/appliance_access_consolecli.py:256:0: traceback: DeprecationWarning: There is no current event loop
.
.
.
```

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -78,6 +78,7 @@ jobs:
           - stable-2.13
           - stable-2.14
           - stable-2.15
+        # - stable-2.16
         # - devel
         # - milestone
     # Ansible-test on various stable branches does not yet work well with cgroups v2.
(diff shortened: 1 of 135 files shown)


alinabuzachis (approved):

mariolenz:

webknjaz:
@mariolenz you should probably skip running the changelog workflow on `push` events and only run in on `pull_request`: https://github.com/ansible-collections/vmware.vmware_rest/commit/d372262c12773521c8a1d5108c39e70e6f3e199a/checks. Also, you could wire it into the main test workflow (via reusable workflows) and connect it to the `check` job, so it could be used in unified gating.

PR #434, opened 2023-11-04
[DNM] 20231104 Re-generate modules with ansible.content_builder

mariolenz:
This is a test, DO NOT MERGE!

```
pip install black==22.3.0 jsonschema jinja2==3.0.3 boto3 baron redbaron ruamel.yaml autoflake
ansible-galaxy collection install git+https://github.com/ansible-community/ansible.content_builder.git
git clone https://github.com/mariolenz/vmware.vmware_rest.git
mkdir -p /collections/ansible_collections/vmware/vmware_rest/tests
cp -r vmware.vmware_rest/tests/sanity /collections/ansible_collections/vmware/vmware_rest/tests
ansible-playbook build.yaml -e manifest_file=MANIFEST.yaml
autoflake --in-place --remove-all-unused-imports /collections/ansible_collections/vmware/vmware_rest/plugins/modules/*.py
cd vmware.vmware_rest/
git checkout -b 20231104-test-ansible.content_builder
rm -rf plugins/modules/
cp -r /collections/ansible_collections/vmware/vmware_rest/plugins/modules/ plugins/
echo -e '---\nminor_changes:\n- "Re-generate modules with ansible.content_builder."' > changelogs/fragments/regenerate_modules.yaml
git add --all
git commit -m "20231104 Re-generate modules with ansible.content_builder"
git push --set-upstream origin 20231104-test-ansible.content_builder
```

build.yaml:
```
---
- hosts: localhost
  gather_facts: yes
  roles:
    - ansible.content_builder.run
```

MANIFEST.yaml:
```
---
collection:
  path: /collections/ansible_collections/vmware/vmware_rest
  namespace: vmware
  name: vmware_rest
plugins:
  - type: module_openapi
    name: "vmware_rest"
    content: cloud
    api_object_path: vmware.vmware_rest/config/api_specifications/7.0.2
    resource: vmware.vmware_rest/config
    action: generate_all
    unique_key: ""
    rm_swagger_json: ""
    module_version: "1.0.0"
    author: "Ansible Cloud Team"
```

PR #435, opened 2023-11-04
[DNM] 20231104 Test ansible.content_builder with 2.16

mariolenz:
This is a test, DO NOT MERGE!

Pretty much the same as #434 but with stable-2.16 sanity tests only.


mariolenz:
@alinabuzachis @GomathiselviS I've already mentioned this in #432:

```
Running sanity test "import" on Python 3.12
Run command with data: importer.py
Run command with data: importer.py
ERROR: Found 132 import issue(s) on python 3.12 which need to be resolved:
ERROR: plugins/modules/appliance_access_consolecli.py:248:0: traceback: DeprecationWarning: There is no current event loop
.
.
.
```

PR #436, opened 2023-11-06
[DNM] 20231106 Re-generate modules with ansible.content_builder

mariolenz:
This is a test, DO NOT MERGE!

Pretty much the same as #434 but with `pip install black` instead of `pip install black==22.3.0`.


mariolenz:
recheck

mariolenz:
recheck

PR #439, opened 2023-11-07
Cleanup tox.ini and GHA

GomathiselviS:
##### SUMMARY

- Ansible-lint pinned to 6.21.x
- Refactored the GitHub workflows so that the files are consistent across all cloud repos.
- Removed references to gouttelette from tox.ini

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


mariolenz:
recheck

mariolenz:
This is weird. It looks like `ansible-test-sanity-docker-stable-2.9` didn't test with 2.9 but with 2.17 the first time. At least, this is how I understand the [log](https://b7cdecd83332b0094d1b-abc20bcf566db6dd6790e38b5800d331.ssl.cf1.rackcdn.com/439/e6576797bd29e522e70765d784e55c1b9b11049c/check/ansible-test-sanity-docker-devel/0c38ac5/job-output.txt):

```
TASK [Install ansible into virtualenv]
controller | Collecting jinja2>=3.0.0 (from ansible-core==2.17.0.dev0)
controller | Collecting PyYAML>=5.1 (from ansible-core==2.17.0.dev0)
controller | Collecting cryptography (from ansible-core==2.17.0.dev0)
controller | Collecting resolvelib<1.1.0,>=0.5.3 (from ansible-core==2.17.0.dev0)
controller | Successfully installed MarkupSafe-2.1.3 PyYAML-6.0.1 ansible-core-2.17.0.dev0 cffi-1.16.0 cryptography-41.0.5 jinja2-3.1.2 pycparser-2.21 resolvelib-1.0.1
```

And:

```
controller | Read 0 sanity test ignore line(s) for Ansible 2.17 from: tests/sanity/ignore-2.17.txt
```

And now `ansible-test-sanity-docker-stable-2.11` seems to have some problem.

mariolenz:
recheck

mariolenz:
recheck

mariolenz:
recheck

mariolenz (commented):
This looks basically OK to me, but ansible-lint GHA fails for some integration tests. I'm not really sure what we should do:

- Fix things
- Make ansible-lint ignore those files
- Remove ansible-lint again (for now)

I tend to the last option so we can merge your other changes. And then work on ansible-lint in a second step. But that's only my opinion and I don't feel very strong on it.

What do you think?

GomathiselviS:
@mariolenz :  We do not run ansible-lint on the integration test targets. So I will add a rule to skip the tests folder.

 With respect to the sanity test failures seen on Zuul, we can ignore them. We will eventually remove the zuul jobs for this collection and move completely to GHA

mariolenz:
> We do not run ansible-lint on the integration test targets. So I will add a rule to skip the tests folder.

:+1: 

> With respect to the sanity test failures seen on Zuul, we can ignore them. We will eventually remove the zuul jobs for this collection and move completely to GHA

Oh, I don't mind those. They didn't look related to your changes and, anyway, the last ansible/check on Zull succeeded. They just confused me and I wanted to document my findings ;-)

mariolenz:
@GomathiselviS ansible-lint failed again, but now with a lot of other issues. I'm afraid you'll have to have another look :-(

mariolenz (commented):
I see you've dropped support for ansible-core < 2.14 now and set the collection version to 2.3.2. I'm not sure if we have guidelines for dropping ansible-core versions, but I know at least _some_ people who consider this a breaking change. Which means a new major release, 3.0.0. Anyway, I think it's definitely not a bugfix. So you should increase at least the minor (2.4.0) and not only the patch version.

Personally, I would drop ansible-core < 2.14 in 3.0.0. That said, I feel a bit unsure about dropping support for ansible-core versions / doing a new major release "just like this". Can't we keep the required ansible (core) version at >=2.9.10 for now and get the GHA workflows running? And _then_ prepare a new major release where we drop ansible-core < 2.14 and maybe do some other things (like completely re-generating everything with ansible.content_builder) properly? This feels a bit rash at the moment.

GomathiselviS:
> I see you've dropped support for ansible-core < 2.14 now and set the collection version to 2.3.2. I'm not sure if we have guidelines for dropping ansible-core versions, but I know at least _some_ people who consider this a breaking change. Which means a new major release, 3.0.0. Anyway, I think it's definitely not a bugfix. So you should increase at least the minor (2.4.0) and not only the patch version.
> 
> Personally, I would drop ansible-core < 2.14 in 3.0.0. That said, I feel a bit unsure about dropping support for ansible-core versions / doing a new major release "just like this". Can't we keep the required ansible (core) version at >=2.9.10 for now and get the GHA workflows running? And _then_ prepare a new major release where we drop ansible-core < 2.14 and maybe do some other things (like completely re-generating everything with ansible.content_builder) properly? This feels a bit rash at the moment.

The upcoming release for this collection will be version 3.0.0. I will update the galaxy file and changelog accordingly. Please note that support for ansible-core 2.13 will be discontinued. Additional details can be found [here](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix).

mariolenz:
> The upcoming release for this collection will be version 3.0.0. I will update the galaxy file and changelog accordingly.

:+1: 

> Please note that support for ansible-core 2.13 will be discontinued. Additional details can be found [here](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix).

Yes, I know ;-)

It's a bit unfortunate because we won't get 3.0.0 into Ansible 9... anyway, let's do it!

I'll prepare a PR to remove the 2.9 / 2.10 / 2.11 / 2.12 Zuul jobs!

Please remember to remove 2.13 from the GHA sanity tests while you're at it. I think it's still mentioned.

mariolenz:
@GomathiselviS 

> I'll prepare a PR to remove the 2.9 / 2.10 / 2.11 / 2.12 Zuul jobs!

ansible/ansible-zuul-jobs#1832

> Please remember to remove 2.13 from the GHA sanity tests while you're at it. I think it's still mentioned.

Please also remove the ignore files 2.9 to 2.23 from tests/sanity/ because we don't need them anymore.

changelogs/fragments/gha_cleanup_version_update.yaml
```diff
@@ -0,0 +1,6 @@
+---
+major_changes:
+- "Remove support for ansible-core 2.13"
```
mariolenz, line 3:
```suggestion
breaking_changes:
- "Remove support for ansible-core < 2.14"
```

According to [Creating a changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment):

> Major changes mean the user can CHOOSE to make a change when they update but do not have to

But users don't have a choice. We're removing support for outdated ansible (core) versions. The only choice they have is to not use 3.0.0. But that's not a problem, breaking changes in a major release are OK ;-)

Additionally, we're not removing support for ansible-core 2.13 only but any version < 2.14. Let's make this clear.

mariolenz:
I think `Sanity tests / Sanity (Ⓐstable-2.16)` fails because of ansible-community/ansible.content_builder#68

plugins/module_utils/vmware_rest.py
```diff
@@ -271,7 +270,7 @@ async def update_changed_flag(data, status, operation):
 
 
 async def list_devices(session, url):
-    existing_entries = []
+    pass
```
jillr, line 273: this seems unnecessary, I would suggest removing this.  if we need to make changes to the generator to do that I'd be ok with merging this for now as it should be harmless and doing that under a separate PR.

jillr (approved):

mariolenz (approved):
LGTM

mariolenz:
recheck

mariolenz:
temp. closing

plugins/module_utils/vmware_rest.py
mariolenz: @jillr I don't think this file is generated. At least, when I generate code with ansible.content_builder (for example here: #441) I don't see it. But I may be wrong... I'm still learning and might have make a mistake somewhere.

jillr: It might not be - I'm more familiar with the generated amazon collection and some of those module_utils lived in an early version of that generator, so calling it out just in case!

mariolenz: @jillr It's already a very big PR with a lot of changes. I suggest to keep it like this and have a closer look at this later ;-)

mariolenz:
@GomathiselviS @jillr Is there anything left to do? Or do you think we can merge when ansible/ansible-zuul-jobs#1832 is merged, too?

jillr:
@mariolenz  I think this one is ready; I have some other zuul PRs to review as well so I will get that one later today

jillr:
@GomathiselviS I think we might need to remove the Depends-On from this ticket, but I'm not certain.

mariolenz:
> I think we might need to remove the Depends-On from this ticket, but I'm not certain.

I did, but still had to merge manually:

> Pull request merge failed: Resource not accessible by integration

I think because the GHA stuff / files in .github, but I'm not 100% sure.

jillr:
> I think because the GHA stuff / files in .github, but I'm not 100% sure.

That's right, thanks!

PR #448, opened 2023-11-17
Disable Codecov

mariolenz:
Disable Codecov for now. I don't think it provides valuable information to us, at least not at the moment.

### DIFF
--- a/.github/workflows/sanity-tests.yaml
+++ b/.github/workflows/sanity-tests.yaml
@@ -111,6 +111,8 @@ jobs:
           # which should improve CI performance. See limitations on
           # https://github.com/ansible-community/ansible-test-gh-action#pull-request-change-detection
           pull-request-change-detection: false
+          # Disable Codecov
+          coverage: never
 
   check:  # This job does nothing and is only used for the branch protection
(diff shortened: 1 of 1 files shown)


mariolenz:
recheck

mariolenz:
@GomathiselviS What do you think about this? I really don't see any valuable information for us from Codcov, it only clutters PRs imho.

GomathiselviS:
> @GomathiselviS What do you think about this? I really don't see any valuable information for us from Codcov, it only clutters PRs imho.
I prefer not to disable Codecov because it offers code coverage feedback directly in pull requests. This helps contributors comprehend how their changes affect test coverage, encouraging them to include tests for their modifications. We might consider implementing a threshold to determine whether the test should pass or fail based on the coverage percentage achieved.

mariolenz:
> I prefer not to disable Codecov because it offers code coverage feedback directly in pull requests. This helps contributors comprehend how their changes affect test coverage, encouraging them to include tests for their modifications. We might consider implementing a threshold to determine whether the test should pass or fail based on the coverage percentage achieved.

OK

PR #443, opened 2023-11-09
Add API 7.0.3 spec

mariolenz:
```
python -m venv --clear ~/get-7.0.3-api-venv
source ~/get-7.0.3-api-venv/bin/activate
pip install pyVmomi
pip install -U https://github.com/vmware/vsphere-automation-sdk-python/archive/refs/tags/v8.0.1.0.tar.gz
python vmsgen.py -vc <vCenter IP> -o <output directory path> --oas 2
deactivate
rm -rf ~/get-7.0.3-api-venv
```

PR #449, opened 2023-11-20
Add 7.0 U3 API spec

mariolenz:
Same as #443, but this time I've used the `--deprecate-slash-rest` and `--fetch-authentication-metadata` flags to generate the specs:

`python vmsgen.py -vc <vCenter IP> -o <output directory path> --oas 2 --deprecate-slash-rest --fetch-authentication-metadata`


mariolenz:
I think the problem are some new requirements that I don't see in the 7.0.2 spec:

- The parameter must be an identifier for the resource type: appliance.monitoring.
- The parameter must be an identifier for the resource type: appliance.networking.interfaces.
- The parameter must be an identifier for the resource type: appliance.services.

If I understand [this function](https://github.com/ansible-community/ansible.content_builder/blob/1a8071e914aa1f25a84670b320c2295736a00818/plugins/action/generate_cloud_modules.py#L150) correct, the fix might be to add some additional mapping:

```
            "appliance.monitoring": "appliance_monitoring_info",
            "appliance.networking.interfaces": "appliance_networking_interfaces_info",
            "appliance.services": "appliance_services_info",
```

mariolenz:
> Am I wrong or we're not running anymore integration tests on Zuul (they are probably not triggered for this change)?

At the moment, this PR only contains the new JSON files. Maybe there were no integration tests because there are no code changes yet?

alinabuzachis (approved):

mariolenz:

PR #489, opened 2024-03-20
Release 3.0.1

machacekondra:
Release 3.0.1


mariolenz (approved):
LGTM

149 changed files is a lot, so take my approval with a grain of salt ;-)

GomathiselviS:
Everything appears good. However, I would prefer this PR to exclusively contain the release-related modifications. Would it be possible to split this PR, moving the module and documentation changes to a separate one? Once the other PR is merged, we can rebase this release PR. @mariolenz @alinabuzachis Your thoughts?

alinabuzachis:
> Everything appears good. However, I would prefer this PR to exclusively contain the release-related modifications. Would it be possible to split this PR, moving the module and documentation changes to a separate one? Once the other PR is merged, we can rebase this release PR. @mariolenz @alinabuzachis Your thoughts?

Any changes to the modules are the reflection of the content_builder's changes. As far as I know, documentation regeneration is part of the prep release PR as a consequence of plugins regeneration. We used to include all these in a prep release PR ( for example: https://github.com/ansible-collections/vmware.vmware_rest/pull/451)

machacekondra:
The relevant changes are in specific commit, but in single PR, if needed it may be split into two PRs. But I think it's not an issue here. The issue here is that we don't which content_builder commit was used to do this release I think that would simplify the review a lot, so I think this is a good time to talk about versioning the content_builder, to make the releases more predictable.

GomathiselviS:
> The relevant changes are in specific commit, but in single PR, so all relevant changes will be nicely seen in git history. But if needed it may be split into two PRs. But I think it's not an issue here. The issue here is that we don't which content_builder commit was used to do this release I think that would simplify the review a lot, because we would have PRs like bump to content_builder 1.0.x. So I think this is a good time to talk about versioning the content_builder, to make the releases more predictable.

My request to split this PR is aimed at enhancing the clarity regarding the intent and scope of these modifications for the stakeholders. While this request is optional, I would prefer to adopt it for future changes. Additionally, including a changelog that outlines how the alterations in content_builder have influenced module and documentation generation would facilitate future reference in the git history. Considering versioning for content_builder is also a valuable aspect that we can address as we proceed with the release process with the current changes.

GomathiselviS (approved):

alinabuzachis (approved):

mariolenz:
Since there are three approvals: Is there any reason to not merge and release this?

PR #498, opened 2024-05-30
fix cloud.common dependency

bardielle:
##### SUMMARY
fix the cloud.common dependency in galaxy.yml

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -12,7 +12,7 @@ tags:
   - vmware
   - virtualization
 dependencies:
-  cloud.common: ">=3.0.0,<4.0.0"
+  cloud.common: ">=3.0.0,<=5.0.0"
 repository: https://github.com/ansible-collections/vmware.vmware_rest.git
 homepage: https://github.com/ansible-collections/vmware.vmware_rest
 issues: https://github.com/ansible-collections/vmware.vmware_rest/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -2,6 +2,7 @@
 profile: production
 
 exclude_paths:
+  - meta/runtime.yml
   - tests/integration
   - tests/sanity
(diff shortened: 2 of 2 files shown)


machacekondra:
Can you please add a description to a commit message why do we need it?

mariolenz (approved):
LGTM

mariolenz:
@alinabuzachis What do you think, should we merge this?

alinabuzachis (approved):

galaxy.yml
```diff
@@ -12,7 +12,7 @@ tags:
   - vmware
   - virtualization
 dependencies:
-  cloud.common: ">=3.0.0,<4.0.0"
+  cloud.common: ">=3.0.0,<=5.0.0"
```
felixfontein, line 15: Shouldn't this be `<5.0.0`, or `<6.0.0`? Otherwise it works fine with 5.0.0, but not with 5.0.1.

bardielle: @alinabuzachis Can we change it to be `cloud.common: >=3.0.0` ?

mariolenz: > Can we change it to be `cloud.common: >=3.0.0` ?

This would also include `cloud.common` 6.0.0 and other major releases. However, a major release introduces breaking changes (that's the definition of a major release) that _might_ affect this collection. So I tend to `>=3.0.0,<6.0.0`.

If it turns out the breaking changes in 6.0.0 don't break anything in this collection, we can always amend this to `>=3.0.0,<7.0.0`. But defining that the collection works with _all_ releases greater 3.0.0 seems a little bit problematic. Like an accident waiting to happen, if you know what I mean ;-)

felixfontein: I would probably change it to `>=3.0.0,<5.0.0`, and then bump again once it becomes clear what kind of breaking changes 5.0.0 contains. Except if you already know that 5.0.0 won't have breaking changes affecting this collection, then `>=3.0.0,<6.0.0` would be best.

Issue #509, opened 2024-08-06
Module documentation contains broken links to other modules


mikemorency:
### Summary

Now that the collection documentation is being generated on PR/MR, we can see that certain modules are referencing other modules that dont exist.

The workflow [here](https://github.com/ansible-collections/vmware.vmware_rest/actions/runs/10268040055/workflow?pr=503#L14) provides the following warnings:
```
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_locallibrary_module.rst:697: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_locallibrary_module.rst:699: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_locallibrary_module.rst:699: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_subscribedlibrary_module.rst:699: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_subscribedlibrary_module.rst:701: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/content_subscribedlibrary_module.rst:701: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.content_library_item_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/vcenter_storage_policies_info_module.rst:125: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.vcenter_storage_policies_module'
/home/runner/work/_temp/docsbuild/head/rst/collections/vmware/vmware_rest/vcenter_vm_hardware_disk_module.rst:309: WARNING: undefined label: 'ansible_collections.vmware.vmware_rest.vcenter_storage_policies_module'
[...]

PR #510, opened 2024-08-06
Mm feature/fix broken doc generation

mikemorency:
##### SUMMARY
This change fixes issues in the module documentation related to bad formatting or broken references. These issues are blocking PRs, and were introduced when we added more workflows to generate docs on each PR

For the RST docs, the issues are simply formatting/escaping problems. Those are easily fixed.
For the module docs with broken references, those references have been removed. A long term fix will need to be figured out since these issues will come back when modules are re-generated.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
content_locallibrary
content_subscribedlibrary
vcenter_storage_policies_info
vcenter_vm
vcenter_vm_hardware_disk
vcenter_vm_power
vcenter_vm_storage_policy

##### ADDITIONAL INFORMATION
I opened an issue to document this problem: https://github.com/ansible-collections/vmware.vmware_rest/issues/509

Ive removed the bad documentation as a quick fix for the current modules but when we generate the next version (which should be vcenter 8.0.2) we will need to either fix the openapi spec before generation or go back through the new modules and fix the docs there.


mikemorency:
The failing checks are due to bad workflow configs already on main (meaning i cant change the "cancelled" results). When this PR is merged, future checks should all pass on main

docs/docsite/rst/dev_guide.rst
```diff
@@ -4,15 +4,15 @@
...
+The Ansible VMware REST collection (on ``Galaxy <https://galaxy.ansible.com/vmware/vmware_rest>``, source code ``repository <https://github.com/ansible-collections/vmware.vmware_rest>``) is maintained by Red Hat and the community.
 
 .. contents::
       :local:
 
 Contribution process
 ====================
 
-The modules of the vmware_rest collection are autogenerated by another tool called `vmware_rest_code_generator <https://github.com/ansible-collections/vmware_rest_code_generator>`.
+The modules of the vmware_rest collection are autogenerated by another tool called ``vmware_rest_code_generator <https://github.com/ansible-collections/vmware_rest_code_generator>``.
```
mariolenz, line 15: This is wrong, isn't it? I'm pretty sure @bardielle did the last release using [ansible.content_builder](https://github.com/ansible-community/ansible.content_builder).

But I think there are some other places where vmware_rest_code_generator is still used. Should we correct this here, or should we concentrate on fixing the broken doc generation and create another PR for this?

I tend to the second option. What do you think?

mikemorency: I think your right, this is out of date.

I agree with your tend-ing. Im supposed to generate specs for vcenter 8 soon so I can document the updated process then

mariolenz (approved):
> The failing checks are due to bad workflow configs already on main (meaning i cant change the "cancelled" results). When this PR is merged, future checks should all pass on main

Well, let's merge the CI and test this?

alinabuzachis (approved):

PR #512, opened 2024-08-12
remove doc workflows

mikemorency:
##### SUMMARY
These workflows validate docs and are supposed to publish documentation to github pages. However, the github pages publish doesnt work correctly and there are a few difficult to solve issues that come up when validating docs. While i like the idea of validating the docs, the difficulty in solving the issues outweighs the benefits in my opinion.

For example, previous validation errors involved links to modules that do not exist. The documentation is pulled from the OpenAPI spec (as are the modules) so the only real option we have is to delete the extra documentation. When we re-generate modules, the errors come back. So we need to do this every time we release.

Additionally, links to other collections do not work without adding that collection as a dependency (as far as i can tell). Links to vmware.vmware are present in the `seealso` block, but these links cant be resolved by these workflows. Im not sure how to fix this.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
GH workflows


mariolenz:
> While i like the idea of validating the docs, the difficulty in solving the issues outweighs the benefits in my opinion.

I'm not sure if I agree. Let's do one or two (or maybe even three) tries to a least make the docs _validation_ work. I'm not sure what the docs publish job really does. So I don't object to dropping this. But it would be great if we could at least make sure that the documentation is OK.

Since this collection is part of the Ansible Community Package (which is a good thing!) you will hear from the Steering Committee if the documentation has any problems. So you'll just postpone problems.

First try: #513

PR #511, opened 2024-08-12
Release 4.0.0

mikemorency:
##### SUMMARY
This is release 4.0.0, which includes new openapi specs for vsphere 8.0.2. It also includes two new modules:
- content_library_info - returns information about an existing content library
- content_library_subscriptions_info - returns information about subscriptions to an existing content library

In addition to the updated modules, the following is included in the release:
1. remove support for ansible-core 2.14
2. Up to date instructions and playbooks for generating module content in this repo. This information has been added to `development.md`
3. various doc updates for lookup plugins and the README

##### ADDITIONAL INFORMATION
Related: https://github.com/ansible-community/ansible.content_builder/pull/89

PR #513, opened 2024-08-12
Add vmware.vmware for docs CI

mariolenz:
##### SUMMARY
Now that some modules reference [vmware.vmware](https://github.com/ansible-collections/vmware.vmware) we see some problems in the CI checking the documentation. I'm not sure if this will fix it, it's just a try.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
.github/workflows/docs-pr.yml

##### ADDITIONAL INFORMATION
#511

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -17,6 +17,7 @@ jobs:
     name: Validate Ansible Docs
     uses: ansible-community/github-docs-build/.github/workflows/_shared-docs-build-pr.yml@main
     with:
+      extra-collections: vmware.vmware
       init-lenient: false
       init-fail-on-error: true
       artifact-name: ${{ github.event.repository.name }}_docs_${{ github.event.pull_request.head.sha }}_validate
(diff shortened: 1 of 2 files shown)


mikemorency:
@mariolenz the modules dont have the `seealso` section on your branch, which is where the error is occuring. That section was added when i re-generated them

mariolenz:
> the modules dont have the `seealso` section on your branch, which is where the error is occuring. That section was added when i re-generated them

I know. But I think even if I add a `seealso` to a module, the CI would use the current CI workflow instead of the changed one from this PR. So I think we have to merge this first and then test again.

Let's give it a try! I mean you can want to drop this completely (#512) and we can always do it. So I don't see any harm in merging this and testing again.

mikemorency:
It's using the ci workflow from your branch. See here 
https://github.com/ansible-collections/vmware.vmware_rest/actions/runs/10356563102/job/28666661588?pr=513
It installs the extra collection you specified

PR #505, opened 2024-08-01
Add GitHub workflows for docs

mikemorency:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #518, opened 2024-08-20
Fix Build Ansible Docs workflow

mariolenz:
##### SUMMARY
Trying to fix the `Build Ansible Docs` CI workflow which fails ATM.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
.github/workflows/docs-push.yml

##### ADDITIONAL INFORMATION
[This](https://github.com/ansible-collections/vmware.vmware_rest/pull/511#issuecomment-2291830581) and the following comment.

### DIFF
--- a/.github/workflows/docs-push.yml
+++ b/.github/workflows/docs-push.yml
@@ -24,6 +24,8 @@ jobs:
       init-fail-on-error: true
       intersphinx-links: |
         ansible_devel:https://docs.ansible.com/ansible-core/devel/
+      provide-link-targets: |
+        ansible_collections.vmware.vmware.vcsa_settings_module
 
   publish-docs-gh-pages:
     # use to prevent running on forks


mikemorency (approved):
thank you for following up. I completely overlooked this

mariolenz:
This fixed the original error, but now there's [a new one](https://github.com/ansible-collections/vmware.vmware_rest/actions/runs/10485967257/job/29043248761).

@mikemorency @alinabuzachis This workflow is quite new, it looks like you've introduced it only 3 weeks ago. Do we really need (or at least want) this? In that case, it would be great if you could make it succeed.

If not, should we simply drop it?

mikemorency:
I got the action to pass, but the github pages link just leads to a 404. I guess ill look into as i have time....

I dont really see any other collection repos with github pages or built RST files. Is this some legacy way of documenting the collections or just a one off quirk?

mariolenz:
I have no idea :shrug: 

You've introduced this in #505 and @alinabuzachis approved. Where did you get this workflow from, and why did you add it?

For community.vmware I've dropped the docs some time ago and simply rely on it being part of the Ansible Community Package and the documentation showing up on [docs.ansible.com](https://docs.ansible.com/ansible/latest/collections/community/vmware/index.html).

Which is also the case for [vmware.vmware_rest](https://docs.ansible.com/ansible/latest/collections/vmware/vmware_rest/index.html).

So if neither you nor @alinabuzachis can give a good reason to have this workflow, I repeat my suggestion to drop it ;-)

Issue #500, opened 2024-06-26
recursive_folder_or_rp_moid_search awaits objects_moid to be in order (but it isn't necessarily)


dbk-rabel:
Hi.

We stumbled across this, because sometimes folder lookups worked fine and sometimes they didn't.

While troubleshooting we noticed that `recursive_folder_or_rp_moid_search` awaits `objects_moid` (which it gets from the API) to be in order of the path.
If they're not, https://github.com/ansible-collections/vmware.vmware_rest/blob/main/plugins/plugin_utils/lookup.py#L472 will never match.

But the API does not always send them in order.

We have a lookup that looks like this:

`lookup('vmware.vmware_rest.folder_moid', '/my_datacenter/my_topfolder/my_subfolder', **connection_args)`

In `get_all_objects_path_moid` there is an API call to a ressource like this:

https://<url>/api/vcenter/folder?datacenters=my_datacenter_moid&names=my_subfolder&names=my_topfolder

And it receives this from the API:

`[{'folder': 'my_subfolder_moid', 'name': 'my_subfolder', 'type': 'VIRTUAL_MACHINE'}, {'folder': 'my_topfolder_moid', 'name': 'my_top_folder', 'type': 'VIRTUAL_MACHINE'}]`

This only happens for folders that we've created in the last days. Not sure if this is the reason, why the API behaves like that.

But since the API is not consistent with the ordering, I think we can say that `recursive_folder_or_rp_moid_search` is broken right now.

Yours
  David

dbk-rabel:
Could this maybe already have been fixed by #445 ?

reusch:
Maybe this is also the same as #324

mikemorency:
hi @dbk-rabel im having some trouble reproducing the issue. Im also experiencing the EmbeddedModuleUnexpectedFailure when I try to use these lookups.

Could you provide the version of the collection your using, the version of ansible packages (ansible and/or ansible-core), and your vcenter version? Also maybe a playbook to reproduce the issue?

I dont see any reason the playbook in 445 wouldnt work for you but other playbooks would, so thats interesting. If you have any insight into why that might be happening that would also be helpful

dbk-rabel:
Hello @mikemorency .

We are using vmware.vmware_rest 3.0.1, ansible-core 2.15.11
vSphere Client version 8.0.2.00200

Here is an example playbook, but of course you would need to create the folders in vcenter first:

```yaml
---
- name: Test
[...]

PR #445, opened 2023-11-14
Fix lookups return empty string when looking up resources inside folders

ThSchrad:
##### SUMMARY
When lookups search for resources inside folders, they return an empty string even if the resource is available in this path. For example, a vm_moid lookup with the term `/mydc/vm/myfolder1/myvm2` returns an empty string while a lookup for `/mydc/vm/myvm1` returns the expected moid.

One could've fixed this issue with no major overhaul of the file. But by looking at the code, I expect that more issues would arise such as finding the wrong folder if multiple folders have the same name, e. g. `/mydc/vm/myfolder1/foo` and `/mydc/vm/myfolder2/foo`.

So I rewrote large parts of the file to strictly follow the path top-down. On some occasions the lookups now make multiple requests against the api to ensure we find the correct resource (e. g. finding the root resource pool on a standalone host). This will certainly decrease performance at the expense of correct results.

Also I think the code is no easier to read and reason about as there is no complex recursive function anymore.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_moid 
datacenter_moid 
datastore_moid 
folder_moid 
host_moid 
network_moid 
resource_pool_moid 
vm_moid 

##### ADDITIONAL INFORMATION
I used an ansible playbook to reproduce the issue. Please note that parts of this playbook (path, moid) must be adjusted to the environment in which the tests are executed. The credentials were set via environment variables.

```yaml
- name: Test
  hosts: localhost
  gather_facts: false
  tasks:
    - name: Test
      ansible.builtin.debug:
        msg: "{{ lookup('vmware.vmware_rest.' + item['type'] + '_moid', item['path']) == item['moid'] }}"
      loop:
        - type: datacenter
          path: /mydc
          moid: datacenter-xxxxxx
        - type: network
          path: /mydc/network/mynet1
          moid: network-xxxxxx
        - type: datastore
          path: /mydc/datastore/myds1
          moid: datastore-xxxxxx
        - type: folder
          path: /mydc/vm
          moid: group-xxxxxx
        - type: folder
          path: /mydc/vm/myfolder1
          moid: group-xxxxxx
        - type: vm
          path: /mydc/vm/myvm1
          moid: vm-xxxxxx
        - type: vm
          path: /mydc/vm/myfolder1/myvm2
          moid: vm-xxxxxx
        - type: cluster
          path: /mydc/host/mycluster1
          moid: domain-xxxxxx
        - type: host
          path: /mydc/host/mycluster1/myhost1
[...]

Issue #324, opened 2022-04-29
folder_moid_lookup not working consistently


robbevl:
Sometimes, a folder 4 levels deep is not found, the lookup then returns an empty string.

Our folderstructure looks like this:
/DC/Applications/MYAPP/PRD

/DC/Applications/MYAPP is found, and MOID returned.
/DC/Applications/MYAPP/PRD is not found

If I create /DC/Applications/MYAPP/PROD it is found.
If that folder is then renamed to /DC/Applications/MYAPP/PRD the plug can't find it any more.

I'm not sure what the issue is:
- It's not the all caps, because it can find /DC/Applications/MYAPP/TESTING
- It's not three letter named, cause it can find /DC/Applications/MYAPP/RND
- It's not even PRD and TST being "special" because it can find /DC/Applications/MYAPP2/PRD. Then again, /DC/Applications/MYAPP2/PROD couldn't be found

It looks totally random, but it's always four levels deep, and once it decides it can't find a name it will _never_ find that name, not even if you delete and recreate the folder, or if you rename a folder that can be found.

Note that [vmware.vmware_rest.vcenter_folder_info](https://github.com/ansible-collections/vmware.vmware_rest/blob/main/docs/vmware.vmware_rest.vcenter_folder_info_module.rst) _can_ find those problematic folders.

If someone can help me pinpoint this I'm all ears.
I'll use a workaround like this for now:

```
- name: Get folder
  vmware.vmware_rest.vcenter_folder_info:
    vcenter_hostname: "{{ vcenter_hostname }}"
    filter_names:
      - "{{ deploy_folder | basename }}"
    parent_folders:
      - "{{ lookup('vmware.vmware_rest.folder_moid', deploy_folder | dirname) }}"
  register: my_folder
```

`"{{ my_folder.value[0].folder }}"` 
gives me then the same value as I expected from
` "{{ lookup('vmware.vmware_rest.folder_moid', deploy_folder) }}"`

robbevl:
While the workaround is working, I noticed that other lookups have the same issue..
I have no choice but convert our deployscript to community.vmware :(

goneri:
Thank you @robbevl for the bug report. What version of vSphere do you use?

goneri:
I've tried to reproduce the problem with the following tasks:

```yaml
- name: Create the Applications folder
  community.vmware.vcenter_folder:
    datacenter_name: my_dc
    folder_name: Applications
    folder_type: vm
[...]

PR #523, opened 2024-08-27
refactor lookup plugins using top down method

mikemorency:
##### SUMMARY
This change rewrites the lookup plugin logic to use a top down approach to finding objects. The plugins will now step through an objects path and try to lookup each intermediate object until it gets to the final object.

The old method has a few issues:
1. The existing plugins use an incorrect search method  which yield inconsistent results (as described in https://github.com/ansible-collections/vmware.vmware_rest/issues/500)
2. They also cannot find resources with the same name, even if the paths given are unique. For example, if `/datacenter/vms/my-folder` and `/datacenter/hosts/my-folder` both exist, you can never find either one
3. Items with the same name will cause conflicts, even if those items have slightly different paths. For example: `/datacenter/vms/foo` and `/datacenter/hosts/foo`  existing at the same time makes it difficult to search for either folder

Fixes:
https://github.com/ansible-collections/vmware.vmware_rest/issues/500
https://github.com/ansible-collections/vmware.vmware_rest/issues/445
https://github.com/ansible-collections/vmware.vmware_rest/issues/324

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
all lookup plugins

##### ADDITIONAL INFORMATION
The original top down approach is described in https://github.com/ansible-collections/vmware.vmware_rest/issues/445. I tried to get that version working but ran into a few issues.


mikemorency:
recheck

changelogs/fragments/523-refactor-lookup-plugins.yml
```diff
@@ -0,0 +1,10 @@
+---
+minor_changes:
+  - cluster_moid - Fix bug where lookup would return incosistent results for objects in nested paths (https://github.com/ansible-collections/vmware.vmware_rest/pull/523)
```
bardielle, line 3: please also mention the issue number.

bardielle (approved):

bardielle:
@mariolenz can you review this PR please?

plugins/plugin_utils/lookup.py
```diff
@@ -12,109 +12,90 @@
...
+                    filter_value
+                )
+            except KeyError:
+                corrected_filters[filter_key] = filter_value
+
+        return corrected_filters
+
+    async def fetch_object_with_filters(self, object_type, filters):
+        _url = self.build_url(object_type, filters)
+        res = await self.fetch(_url)
```
machacekondra, line 89: why twice?

plugins/plugin_utils/lookup.py
```diff
@@ -12,109 +12,90 @@
...
+        self.session = session
+
+    async def fetch(self, url):
+        async with self.session.get(url) as response:
+            result = await response.json()
+            return result
+
+    def build_url(self, object_type, filters):
+        corrected_filters_for_query = self.correct_filter_names(filters, object_type)
+        if object_type == "resource_pool":
```
machacekondra, line 57: Maybe worth to make it generic?
Like `if '_' in object_type`?

machacekondra (approved):

plugins/plugin_utils/lookup.py
mikemorency: i think the old implementation of the plugin used both methods. Theres no reason for two different methods now though, ill combine them

plugins/plugin_utils/lookup.py
mikemorency: good point, i think theres no reason to even have the `if` statement too

PR #524, opened 2024-08-27
Drop scenario guide

mariolenz:
##### SUMMARY
Let's drop the scenario guide. It's not really maintained and tends to make problems.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
docs/docsite/extra-docs.yml

PR #525, opened 2024-08-28
release 4.0.1

mikemorency:
##### SUMMARY
There was an issue uploading release 4.0.0 to RedHat Automation Hub, so they've asked us to create a new release.

There is no functionality difference between 4.0.0 and 4.0.1.

##### ISSUE TYPE
- Release

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,16 @@
 =================================
-vmware.vmware_rest Release Notes
+vmware.vmware\_rest Release Notes
 =================================
 
 .. contents:: Topics
+
+v4.0.1
+======
+
+Minor Changes
+-------------
+
+- Removed the scenario guides which are pretty much unmaintained and, therefor, possibly outdated and misleading (https://github.com/ansible-collections/vmware.vmware_rest/pull/524).
 
 v4.0.0
 ======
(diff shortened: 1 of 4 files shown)


mariolenz (changes requested):
A minor change means we need a minor release (4.1.0), a bugfix release (4.0.1) would break semantic versioning ;-)

I see two possibilities:

1. Consider if we can categorize #524 as a bugfix (because the scenario guides are old and buggy) and update the two changelog files accordingly
2. Prepare 4.1.0 instead of 4.0.1

I _tend_ to 2 but am not 100% sure.

mikemorency:
hmmm ok. technically documentation changes dont even need a changelog, so i think keeping this as a bugfix release instead of a minor release would be more accurate. ill update the changelogs from 524

CHANGELOG.rst
```diff
@@ -1,9 +1,17 @@
...
+vmware.vmware\_rest Release Notes
 =================================
 
 .. contents:: Topics
 
+v4.0.1
+======
+
+Minor Changes
+-------------
```
mariolenz, line 11:
```suggestion
Bugfixes
--------
```

It's a little bit weird if `changelog/changelog.yaml` says it's a bugfix and `CHANGELOG.rst` says it's a minor change ;-)

Apart from this LGTM!

mikemorency:
@mariolenz can you approve/remove the change request if it looks good? I think its blocking the merge automation

mariolenz (approved):
LGTM

PR #529, opened 2024-09-10
Mm feature/add lookup tests

mikemorency:
##### SUMMARY
Including a playbook that can be used to test the lookup plugins. Its designed to be run manually and needs inputs from a specific environment.

Once automated testing is set up, this test can be modified to require less inputs from the user

Also included a bugfix which was found while testing. Lookng up the contents of a datacenter would raise an exception instead of returning the correct results

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Lookup plugins

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -5,6 +5,7 @@ exclude_paths:
   - meta/runtime.yml
   - tests/integration
   - tests/sanity
+  - tests/manual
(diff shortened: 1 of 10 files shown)


anna-savina (approved):
LGTM

bardielle (approved):

PR #538, opened 2024-10-09
Add eco vcenter upstream ci capabilities

shellymiron:
##### SUMMARY
This PR aims to insert self hosted runner to enable integration tests to run on the eco vcenter environment.
This includes several adjustments to the code and additional action to be triggered for each PR, and run all the
integration tests on that env with the PR's new code.


mikemorency:
the mergeit label will try to merge the pr AFAIK. it should be added once someone reviews

shellymiron:
recheck

shellymiron:
recheck

shellymiron:
recheck

shellymiron:
@anna-savina @mikemorency hey! please review :) this PR insert the eco ci job capabilities without editing a specific integration test - once a PR to adjust a test will be open, the job will catch it and run it by the the prefix of the test - `vmware_rest `

mikemorency:
I was thinking that the test env will need to install the `requirements.txt` in the root of the repo. But then I looked at vmware.vmware and it doesnt look like we install any python requirements for those tests. So Im not really sure how it works but I guess we dont need to install anything extra for these tests too?

mikemorency (approved):

shellymiron:
@mikemorency yeah I looked at it aswell and assumed we don't need to install any requirements - worst case scenario is to add it later as a step in the Makefile if we will need it for the tests

anna-savina (approved):
LGTM

PR #543, opened 2024-10-28
Add eco-vcenter-ci workflow

shellymiron:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,46 @@
+---
+name: Ansible Eco vCenter Integration Test
+on:
+  pull_request_target:
+    types: [opened, synchronize]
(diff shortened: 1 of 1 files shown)


mariolenz:
@shellymiron Since you've closed this PR, I think you should also close the `eco-vcenter-ci-patch-1` if it isn't needed anymore / you don't plan to use or recycle it form something else. Otherwise, things might become a little bit confusing.

For example, the are some `patchback/$SOMETHING` branches that confuse me. I just don't know if they are leftovers or if we still need them ;-)

PR #545, opened 2024-11-04
Enhance appliance_health target

prabinovRedhat:


tests/integration/targets/vmware_rest_appliance/tasks/appliance_health.yml
```diff
@@ -0,0 +1,157 @@
+---
+- name: Get the database health status
+  vmware.vmware_rest.appliance_health_database_info:
+    vcenter_validate_certs: "{{ VMWARE_VALIDATE_CERTS }}"
+    vcenter_hostname: "{{ VMWARE_HOST }}"
```
anna-savina, line 5: if credentials passed as env var, it should work without additional assignment. No need specify hostname, username and password.

tests/integration/targets/vmware_rest_appliance/tasks/appliance_health.yml
```diff
@@ -0,0 +1,157 @@
+---
+- name: Get the database health status
```
anna-savina, line 2: this task is failing, see test results

tests/integration/targets/vmware_rest_appliance/tasks/appliance_health.yml
prabinovRedhat: So, we need to fix it and pass it as environment variables. Do you want to open a task for it? Or add it in my PR as well?

tests/integration/targets/vmware_rest_appliance/tasks/appliance_health.yml
prabinovRedhat: Yes, I know. I think that’s why ignore_errors: true was set in the original test.

tests/integration/targets/appliance/tasks/main.yml
```diff
@@ -2,7 +2,6 @@
 - import_tasks: appliance_access_dcui.yml
 - import_tasks: appliance_access_shell.yml
 - import_tasks: appliance_access_ssh.yml
-- import_tasks: appliance_health.yml
```
shellymiron, line 5: maybe for now don't delete this test from the appliance, we will deal with zuul later

shellymiron (approved):

tests/integration/targets/appliance/tasks/main.yml
prabinovRedhat: I mean the existing tests aren’t running on Zuul

shellymiron: yes, however, if for some reason we would like to enable them, we need them to exists still IMO
when we will finish all the test cases modification, we will think on deletion

mikemorency (approved):

PR #550, opened 2024-11-15
fix nonsensical path results in lookup plugin

mikemorency:
##### SUMMARY
If someone specifies a path to an object (like a cluster) that cannot contain the lookup object, the plugin ignores the cluster name and just uses the datacenter as a filter.

Ideally, I think we would throw an error telling the user that they are searching for a path the doesnt make sense. But the old plugin output is just an empty string, so I am keeping that output.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
lookup plugins

### DIFF
--- a/plugins/plugin_utils/lookup.py
+++ b/plugins/plugin_utils/lookup.py
@@ -17,6 +17,16 @@ from ansible_collections.vmware.vmware_rest.plugins.module_utils.vmware_rest imp
     open_session,
 )
 
+
+class InvalidVspherePathError(Exception):
+    def __init__(self, container_name, object_type):
+        self.container_name = container_name
+        self.object_type = object_type
+        super().__init__(
+            "VSphere object '%s' cannot hold objects of type '%s'" % (container_name, object_type)
+        )
+
+
 FILTER_MAPPINGS = {
     "resource_pool": {
         "parent_resource_pools": "parent_resource_pools",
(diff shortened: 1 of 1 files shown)


anna-savina (approved):
LGTM

plugins/plugin_utils/lookup.py
```diff
@@ -185,7 +199,7 @@ async def process_intermediate_path_part(self, intermediate_object_name):
                 self.set_new_filters_with_datacenter({"resource_pools": result})
                 return result
 
-        if self.object_type in ("host", "resource_pool"):
+        if self.object_type in ("vm", "host", "resource_pool"):
```
bardielle, line 202: Why did you added the "vm"? it's not clear 
If it is VM it will get inside the previous VM

bardielle: The function become not clear what each "if" statement comes to check and return 
can you refactoring this long function?

mikemorency: The vm, host, and resource pool objects accept a cluster name as part of their filter spec. https://vdc-download.vmware.com/vmwb-repository/dcr-public/1cd28284-3b72-4885-9e31-d1c6d9e26686/71ef7304-a6c9-43b3-a3cd-868b2c236c81/doc/structures/com/vmware/vcenter/vm/svc.filter_spec-structure.html
So this change is fixing that

mikemorency: I refactored the method and renamed it to `add_intermediate_path_part_to_filter_spec`. I also added some more comments and documentation, hopefully its clearer now

bardielle (approved):

Issue #540, opened 2024-10-10
content_library_item_info fails when you only specify library_id


mikemorency:
### Summary

The content_library_item_info module fails when a user specifies just the library_id. This triggers the 'list' method, which has an incorrect API url. Heres the relevant lines: https://github.com/ansible-collections/vmware.vmware_rest/blob/main/plugins/modules/content_library_item_info.py#L292-L298

As far as I can tell, the module generation is working as expected. Heres the VMware source docs: 
https://developer.broadcom.com/xapis/vsphere-automation-api/latest/content/api/content/library/item__library_id/get/

It currently is:
```
    return yarl.URL(
        ("https://{vcenter_hostname}" "/api/content/library/item?library_id").format(
            **params
        )
        + gen_args(params, _in_query_parameters),
        encoded=True,
    )
```

It should be:
```
    return yarl.URL(
        ("https://{vcenter_hostname}" "/api/content/library/item").format(
            **params
        )
        + gen_args(params, _in_query_parameters),
        encoded=True,
    )
```

Obviously the quick fix is a simple change. However, since the module will just be re-generated with the incorrect URL I'd like to discuss how best to fix this long term. 

We could migrate the module functionality to vmware.vmware, but whats the best way to communicate this to users in this collection? Would we be able to deprecate this module with a minor release, and then remove it with a major release down the line? Is that the best course of action?

### Issue Type

Feature Idea

### Component Name

content_library_item_info

### Additional Information

```
---
- name: Test Playbook
  hosts: localhost
  gather_facts: false
  tasks:
    - name: Foo
      vmware.vmware_rest.content_library_item_info:
        vcenter_validate_certs: false
        library_id: "{{ my_library_id }}"
      register: ds_lib_items_info
    - debug:
        var: ds_lib_items_info
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

mikemorency:
@mariolenz @bardielle Id like your opinions on this please

asulliv1:
I just spend a good hour trying to figure this problem out.  after I figure out the code change I then found this issue..

I have been getting to following error "Unsupported property with name: library_id?library_id."

I was able to verify that modifying the code to 

```
    return yarl.URL(
        ("https://{vcenter_hostname}" "/api/content/library/item").format(
            **params
        )
        + gen_args(params, _in_query_parameters),
        encoded=True,
    )
```
Allowed me to pull all content library items.   

[...]

PR #551, opened 2024-11-19
deprecate content_library_item_info due to module generation issues

mikemorency:
##### SUMMARY
Fixes issue
https://github.com/ansible-collections/vmware.vmware_rest/issues/540

Deprecate module content_library_item_info in favor of vmware.vmware.content_library_item_info. The module in this collection was not generated correctly, leading to it failing in some use cases. We cannot easily fix this since the issue is in the openapi spec provided by vmware.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
content_library_item_info

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -25,6 +25,7 @@ jobs:
         ansible_devel:https://docs.ansible.com/ansible-core/devel/
       provide-link-targets: |
         ansible_collections.vmware.vmware.vcsa_settings_module
+        ansible_collections.vmware.vmware.content_library_item_info_module
 
   build-docs:
     permissions:
@@ -39,6 +40,7 @@ jobs:
         ansible_devel:https://docs.ansible.com/ansible-core/devel/
       provide-link-targets: |
         ansible_collections.vmware.vmware.vcsa_settings_module
+        ansible_collections.vmware.vmware.content_library_item_info_module
 
   # This job requires that the workflow trigger is pull_request_target. However that trigger uses
   # whatever version of the workflow is available on main instead of the feature branch.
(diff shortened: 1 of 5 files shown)


billy-stellar:
@mikemorency does this mean vmware.vmware will soon have a .content_library_item_info module? I don't see that documented today, nor in the source code.

mikemorency:
Hi @billy-stellar , 

Yes, we will be adding a content_library_item_info module with the next minor release. The PR to add the module is open in the vmware.vmware repo.
Both vmware.vmware and vmware.vmware_rest have minor releases planned in the next ~2 weeks

mariolenz:
@mikemorency Some of the tests fail. I think I had a similar problem in `community.vmware` and solved it like this:

https://github.com/ansible-collections/community.vmware/blob/394a5a6bc05288af8054e8d1e20cba94a2957b7d/.github/workflows/extra-docs-linting.yml#L34-L37

I'm not sure if you would also have to install `vmware.vmware` somewhere explicitly, but I think you should wait until the module really is part of `vmware.vmware` to make the CI happy.

> Both vmware.vmware and vmware.vmware_rest have minor releases planned in the next ~2 weeks

There will be new releases of the [Ansible Community Package](https://pypi.org/project/ansible/) December 3rd, and the next one will be [end of January](https://forum.ansible.com/t/10988). It would be great if you could do those releases in time so we can include them! Of course, this will only work if we're talking about minor releases here ;-)

BTW those will be the last 9.x and 10.x releases. From then on, it's ansible 11 or nothing.

mariolenz (approved):
LGTM

Thanks @mikemorency!

mariolenz (approved):
LGTM

mariolenz:
> Pull request merge failed: Resource not accessible by integration, You may need to manually rebase your PR and retry.

@mikemorency I think `mergeit`doesn't work if you there are are any changes to the workflows. Let's just merge it the hard way and test afterwards.

mariolenz:
It looks like the integration tests still succeed: #553

So let's say it's OK :-)

Issue #385, opened 2023-01-23
vmware.vmware_rest.vcenter_ovf_libraryitem - Update existing template


sepetuks:
##### SUMMARY

I am trying to update existing OVF template but it looks it is not updating it when I provide library_item_id.
All works ok when I need first OVF template to be cloned to library but I do not succed updateing Connten library template with new versions.
As per documentation"
`- library_item_id (str)`: Identifier of the library item that should be should be **updated**. (['present'])
https://github.com/ansible-collections/vmware.vmware_rest/blob/main/docs/vmware.vmware_rest.vcenter_ovf_libraryitem_module.rst
Per documentation I read it should be possible. So I either thats issue with documentation or with code.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

vmware.vmware_rest.vcenter_ovf_libraryitem

##### ANSIBLE VERSION

```paste below
ansible [core 2.13.7]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/henris/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/henris/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/henris/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/henris/.local/bin/ansible
  python version = 3.8.10 (default, Nov 14 2022, 12:59:47) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection         Version
------------------ -------
vmware.vmware_rest 2.2.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

vCenter 7.0 Update 3j (20990077)

##### STEPS TO REPRODUCE

```yaml
    - name: Export the VM as an OVF on the library (new)
      vmware.vmware_rest.vcenter_ovf_libraryitem:
        vcenter_hostname: "{{ vcenter_hostname }}"
        vcenter_username: "{{ vcenter_username }}"
        vcenter_password: "{{ vcenter_password }}"
        vcenter_validate_certs: false
        vcenter_rest_log_file: '/tmp/rest.log'
        session_timeout: 2900
        source:
          type: VirtualMachine
          id: '{{ vm1_info.id }}'
        target:
          library_id: '{{ library_id }}'
        create_spec:
          name: golden
          description: an OVF example
          flags: []
        state: present
[...]

PR #560, opened 2024-12-02
update vcenter_ovf_libraryitem docs

mikemorency:
##### SUMMARY
Related to the discussion in https://github.com/ansible-collections/vmware.vmware_rest/issues/385

This change updates the module documentation to mention that template metadata is not updated as expected. It also includes a workaround so users can update the template description if needed.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
vcenter_ovf_libraryitem


plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -261,93 +265,85 @@
...
 
-- name: Create a new VM from the OVF and specify the host and folder
+- name: Update The Template Description
   vmware.vmware_rest.vcenter_ovf_libraryitem:
-    ovf_library_item_id: '{{ (lib_items.value|selectattr("name", "equalto", "golden_image")|first).id }}'
-    state: deploy
+    session_timeout: 2900
+    source:
+      type: VirtualMachine
+      id: "{{ lookup('vmware.vmware_rest.vm_moid', vm_name) }}"
```
anna-savina, line 309: for consistency
```suggestion
      id: "{{ lookup('vmware.vmware_rest.vm_moid', 'My VM') }}"
```

plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -261,93 +265,85 @@
...
     target:
-      resource_pool_id: "{{ lookup('vmware.vmware_rest.resource_pool_moid', '/my_dc/host/my_cluster/Resources') }}"
-      folder_id: "{{ lookup('vmware.vmware_rest.folder_moid', '/my_dc/vm') }}"
-      host_id: "{{ lookup('vmware.vmware_rest.host_moid', '/my_dc/host/my_cluster/esxi1.test/test_vm1') }}"
-    deployment_spec:
-      name: my_vm_from_ovf_on_a_host
-      accept_all_EULA: true
-      storage_provisioning: thin
+      library_id: "{{ library_id }}"
```
anna-savina, line 311: for consistency
```suggestion
      library_id: 1111111-1111111-1111111-111111
```

plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -261,93 +265,85 @@
...
+      library_item_id: "{{ _ovf_item.id }}"
```
anna-savina, line 312: according to the real result
```suggestion
      library_item_id: "{{ _ovf_item.ovf_library_item_id }}"
```

plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -261,93 +265,85 @@
...
-    error:
-      errors: []
-      information: []
-      warnings: []
-    resource_id:
-      id: vm-1078
-      type: VirtualMachine
-    succeeded: 1
   type: dict
+  sample:
```
anna-savina, line 331: I got this result with `vmware.vmware.content_library_item_info` module, but `vmware.vmware_rest.vcenter_ovf_libraryitem` module returns:
```
    "_ovf_item": {
        "ansible_facts": {
            "discovered_interpreter_python": "/usr/bin/python3"
        },
        "changed": true,
        "failed": false,
        "value": {
            "error": {
                "errors": [],
                "information": [],
                "warnings": []
            },
            "ovf_library_item_id": "ea695665-307d-44da-b813-398969fc3903",
            "succeeded": true
        }
    }
```

plugins/modules/vcenter_ovf_libraryitem.py
mikemorency: `ovf_library_item_id` is only returned if the item already exists. If the item does not exist, the `id` attribute is returned. So i think the original is correct in this example

plugins/modules/vcenter_ovf_libraryitem.py
anna-savina: hm... Are you sure? Update task gave me the same result as Create task. Can you check it please one more time

mikemorency: I updated the sample block. Hopefully its clearer that there are two possible return values

anna-savina (approved):
LGTM

PR #561, opened 2024-12-02
update references to deprecated library item info module

mikemorency:
##### SUMMARY
This change updates references to the deprecated content_library_item_info module. It changes the collection from vmware.vmware_rest to vmware.vmware

I updated the tests and examples. I don't think we should update the module parameter descriptions since those are autogenerated. The links in those descriptions go the the main module page, which mentions that module is deprecated and users should use the new version anyway.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
content_library_item_info


anna-savina:
Should also be add `requirements.yml` with `vmware.vmware` collection?

anna-savina:
> I don't think we should update the module parameter descriptions since those are autogenerated.
This doesn't seem right to me, the documentation looks confusing. Is there any other way to change it?

mikemorency:
> Should also be add `requirements.yml` with `vmware.vmware` collection?

Im not sure which requirements.yml your referring to, but the integration tests one already has vmware.vmware. And vmware.vmware is not a requirement for using this collection, even though its referenced in the docs, so I dont think it should be added to the galaxy.yml dependencies.

PR #572, opened 2024-12-23
modify library ovf clone test to run on the eco vsphere env

shellymiron:
This test takes between 3 to 5 minutes due to OVF image creation, which is not too long (it was much more then that before the modification), but it can still make the PRs a little slower to get merge, so FYI.

The purpose of the test is to verify few vmware_rest modules related to ovf such as `content_locallibrary_info` `vcenter_ovf_libraryitem`. 
The modules meant to enable the user to create vms from OVF image on a specific content library etc...

**Note**: seems like for these modules we need to add `requirements.txt` file with python packages to be installed, so I added it aswell in this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,10 @@ prepare_symlinks:
 install-ansible-collections:
 	ansible-galaxy collection install --upgrade -r tests/integration/requirements.yml
 
+.PHONY: install-python-packages
+install-python-packages:
+	pip3 install -r tests/integration/requirements.txt
+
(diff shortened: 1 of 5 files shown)


mikemorency:
i wonder what the default disk size is for these new vms? maybe we could make that smaller to speed up the ovf export since were not booting the vm

shellymiron:
@mikemorency IIUC, 512MB is the minimal required disk size and that is what we are using

mikemorency (approved):

PR #573, opened 2024-12-27
make sure prepare_lab is only run on zuul

mikemorency:
##### SUMMARY
Adding a check to this test so it cant be run against the Eco vCenter. The first set of tasks in this test poweroff all vms and delete things, so it should only be run on Zuul where that is OK.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
prepare_lab


tests/integration/targets/prepare_lab/tasks/main.yml
```diff
@@ -1,3 +1,10 @@
+---
+- name: Fail If Running On Eco vCenter
+  ansible.builtin.fail:
+    msg: This test is very desctuctive and should only be run on Zuul!
+  when: >-
+    lookup('ansible.builtin.env', 'VCENTER_HOSTNAME') == "eco-vcenter-server.lab.eng.tlv2.redhat.com"
```
shellymiron, line 6: let's put this domain (`eco-vcenter-server.lab.eng.tlv2.redhat.com`) inside a var so that if in the future the env will change we could change it directly from the var. wdyt?

mikemorency: i wasnt really sure where this role gets its vars from, i think theres some static file on the zuul infra somewhere. but i added it to the `prepare_lab` role vars

shellymiron: ah ok, that's got aswell

shellymiron:
I see the tests are failing on the appliance network, this was commented by @prabinovRedhat and a bug opened
so just need to rebase the PR in order for the tests to pass

anna-savina (approved):

shellymiron (approved):

PR #576, opened 2024-12-29
Bugfix/report changes in filesystem directories

OrrBG:
##### SUMMARY
Fixes: fixed filesystem directories module return value

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module_utils


plugins/module_utils/vmware_rest.py
```diff
@@ -203,6 +203,9 @@ async def update_changed_flag(data, status, operation):
     elif operation == "delete" and status in [200, 204]:
         data["failed"] = False
         data["changed"] = True
+    elif operation in ["create", "move", "create_temporary"] and status in [204]:
```
mikemorency, line 206:
```suggestion
    elif operation in ["create", "move"] and status in [204]:
```
if you look at the API spec for this module, the `create_temporary` action never returns a 204. `delete` does return a 204 but that is handled by the previous if statement, so we dont need to include it here
https://raw.githubusercontent.com/ansible-collections/vmware.vmware_rest/refs/heads/main/config/api_specifications/8.0.2/vcenter.json
search for 'directories?action' to find the parts im talking about

mikemorency: you may want to add `create_temporary` to one of the other if statements that checks for 200

OrrBG: Wow! That link is very useful, thanks

mikemorency (approved):

PR #579, opened 2025-01-09
Modified appliance_services target

prabinovRedhat:
Modified appliance_services test to run on real vCenter env

### DIFF
--- a/tests/integration/targets/vmware_rest_appliance/tasks/appliance_services.yml
+++ b/tests/integration/targets/vmware_rest_appliance/tasks/appliance_services.yml
@@ -1,0 +1,100 @@
+---
+- name: Manage and test ntpd service
+  block:
+    - name: Get information about the ntpd service
+      vmware.vmware_rest.appliance_services_info:
+        service: ntpd
+      register: initial_ntpd_info
+
+    - name: Debug initial ntpd service info
+      ansible.builtin.debug:
+        var: initial_ntpd_info
+
+    - name: Stop the ntpd service
+      vmware.vmware_rest.appliance_services:
+        service: ntpd
+        state: stop
+      register: stop_result
+
+    - name: Debug stop result
+      ansible.builtin.debug:
+        var: stop_result
+
+    - name: Get information about the ntpd service after it is stopped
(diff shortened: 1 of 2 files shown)


tests/integration/targets/vmware_rest_appliance/tasks/appliance_services.yml
```diff
@@ -0,0 +1,112 @@
...
+          - after_restart_ntpd_info is defined
+          - after_restart_ntpd_info.failed == false
+          - after_restart_ntpd_info.value.state is defined
+          - after_restart_ntpd_info.value.state == "STARTED"
+        fail_msg: "Failed to restart the ntpd service."
+
+  rescue:
+    - name: Handle errors during ntpd service management
+      ansible.builtin.debug:
+        msg: "An error occurred while managing the ntpd service. Please check the logs for more details."
```
anna-savina, line 100: So the test will never fail and to find out if the test failed, you need to check the logs even after a successful run? It doesn't look very right. No one will check the logs of a test that was successful.

tests/integration/targets/vmware_rest_appliance/tasks/appliance_services.yml
```diff
@@ -0,0 +1,112 @@
...
+
+  always:
+    - name: Ensure ntpd service is started
+      vmware.vmware_rest.appliance_services:
+        service: ntpd
+        state: start
+      when: after_restart_ntpd_info.value.state is defined and after_restart_ntpd_info.value.state != "STARTED"
```
anna-savina, line 107: Better to start it any way even if `after_restart_ntpd_info` is not defined

prabinovRedhat: fixed

tests/integration/targets/vmware_rest_appliance/tasks/appliance_services.yml
prabinovRedhat: fixed

anna-savina (approved):
LGTM

PR #580, opened 2025-01-12
remove redundant libraryitems example modules

OrrBG:
##### SUMMARY
remove redundant vcenter_vmtemplate_libraryitems example modules

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
vcenter_vmtemplate_libraryitems


plugins/modules/vcenter_vmtemplate_libraryitems.py
```diff
@@ -392,15 +346,6 @@
...
 
-- name: Get the list of items of the NFS library
-  vmware.vmware.content_library_item_info:
-    library_id: '{{ nfs_lib.id }}'
-  register: lib_items
-
-- name: Use the name to identify the item
-  set_fact:
-    my_template_item: "{{ lib_items.value | selectattr('name', 'equalto', 'golden-template')|first }}"
-
```
mikemorency, line 403: these tasks are needed for the next task. `my_template_item` is used. 

itd be better if you could just use the output from the first time `vcenter_vmtemplate_libraryitems` is called, but I am not sure thats possible. Youll have to test it out

mikemorency (changes requested):

plugins/modules/vcenter_vmtemplate_libraryitems.py
OrrBG: I tested it out and it seems to work the same whether I use `my_template_item` or `mylib_item`. Furthermore, I printed both variables and observed that they have an identical ID value so this should be fine

plugins/modules/vcenter_vmtemplate_libraryitems.py
```diff
@@ -348,24 +348,6 @@
     memory:
       hot_add_enabled: true
       size_MiB: 1024
-    disks:
```
anna-savina, line 351: Also fix the return value according to the new VM parameter (just one default disk and no nics), please.
Should be something similar:
```
            "cpu": {
                "cores_per_socket": 1,
                "count": 1
            },
            "disks": {
                "2000": {
                    "capacity": 17179869184,
                    "disk_storage": {
                        "datastore": "datastore-3036"
                    }
                }
            },
            "guest_OS": "RHEL_7_64",
            "memory": {
                "size_MiB": 1024
            },
            "nics": {},
            "vm_home_storage": {
                "datastore": "datastore-3036"
            },
            "vm_template": "vm-77122"
```

anna-savina (approved):

mikemorency (approved):

PR #584, opened 2025-02-24
Modified vcenter_vm_scenario1 test target to run on the eco vsphere env

anna-savina:
Modified vcenter_vm_scenario1 test target to run on the eco vsphere env


tests/integration/targets/vmware_rest_vcenter_vm_scenario/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Run vmware_rest_vcenter_vm_scenario test
+  block:
+    - name: Import eco-vcenter common vars
+      ansible.builtin.include_vars:
+        file: ../group_vars.yml
+      tags: eco-vcenter-ci
+
+    - name: Set facts for tests
```
mikemorency, line 9: I think we should add an assertion after this with a helpful error message if any of these are null. It would be good to indicate if theres a module failure, or if the test environment is somehow messed up

tests/integration/targets/group_vars.yml
```diff
@@ -4,7 +4,8 @@
 vcenter_cluster_name: "Eco-Cluster"
 vcenter_datacenter: "Eco-Datacenter"
 vcenter_port: 443
-rhel_9_3_iso_path: "[eco-nfs-datastore-iso] rhel-9.3-x86_64-dvd.iso"
+rhel_9_3_iso_path: "[eco-nfs-datastore-iso] rhel-9.3-x86_64-boot.iso"
```
mikemorency, line 7: i think tzach made a content library with these ISOs that might be better long term than the NFS datastore.
Theres also a content library dedicated to testing resources I set up for vmware.vmware, ansible_ci_resources or something like that. We could add these ISOs there

tests/integration/targets/vmware_rest_vcenter_vm_scenario/runme.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+source ../init-eco.sh
+
+# Generates a string starting with 'test-' followed by 4 random lowercase characters
+TINY_PREFIX="test-$(uuidgen | tr -d '-' | cut -c1-4 | tr '[:upper:]' '[:lower:]')"
```
mikemorency, line 5:
```suggestion
TINY_PREFIX="test-vmware_rest-$(uuidgen | tr -d '-' | cut -c1-4 | tr '[:upper:]' '[:lower:]')"
```
Since we have 3 repos making test resources, it might be helpful to differentiate the prefix so we know where resources are coming from

tests/integration/targets/vcenter_vm_scenario1/refresh_RETURN_block.sh
mikemorency, line 1: this file is referenced here: https://github.com/ansible-collections/vmware.vmware_rest/blob/main/development.md#optional-update-the-return-block-of-the-vmware-modules-using-the-test-suite
I think we can delete that whole section of the docs. It does more harm than good and now this file is being removed

anna-savina: I'll add to the new PR to clean up all outdated tests

tests/integration/targets/vmware_rest_vcenter_vm_scenario/tasks/main.yml
anna-savina: Added to verify_env_information.yml

mikemorency (approved):

PR #598, opened 2025-05-16
bump cloud common ver

mikemorency:
##### SUMMARY
Bumps the cloud.common dependency version to include the latest 2.19 support changes

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -12,7 +12,7 @@ tags:
   - vmware
   - virtualization
 dependencies:
-  cloud.common: ">=3.0.0,<=5.0.0"
+  cloud.common: ">=4.0.0"
 repository: https://github.com/ansible-collections/vmware.vmware_rest
 homepage: https://github.com/ansible-collections/vmware.vmware_rest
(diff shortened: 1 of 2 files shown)


galaxy.yml
```diff
@@ -12,7 +12,7 @@ tags:
   - vmware
   - virtualization
 dependencies:
-  cloud.common: ">=3.0.0,<=5.0.0"
+  cloud.common: ">=4.0.0"
```
mariolenz, line 15:
```suggestion
  cloud.common: ">=4.1.0,<5.0.0"
```

In order to get the fixes for ansible-core 2.19, `>=4.0.0` isn't enough since they've been released in 4.1.0. And it wouldn't be what you've written into the changelog ;-)

Additionally, I suggest to also restrict the version with `<5.0.0` in order to avoid possible problems with breaking changes that 5.0.0 might introduce. If this collection doesn't run into problems with 5.0.0, it's easy and not a breaking change to make this `<6.0.0` later.

mariolenz (approved):
LGTM

Thanks @mikemorency!

anna-savina (approved):

PR #599, opened 2025-05-16
refactor integration tests to be easier to call

mikemorency:
##### SUMMARY
This refactors the integration tests to match vmware.vmware and work more seamlessly with ansible-test. It also simplifies how developers can call the tests.

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION
We still had some tests that are not run at all. I think they are left over from the zuul testing. I went ahead and removed them. but let me know if I am wrong


tests/integration/targets/vmware_rest_appliance-multi-hosts/tasks/main.yml
```diff
@@ -1 +1,10 @@
-- import_tasks: appliance_ntp.yml
+- name: Test
+  # https://github.com/ansible-collections/cloud.common/issues/179
+  environment:
```
anna-savina, line 3:
```suggestion
  environment: "{{ environment_auth_vars }}"
```

docs/maintainer.md
```diff
@@ -34,8 +34,3 @@ When you are ready to add a new version of Ansible core, you need to add it to t
 Testing is done as a part of the pull request workflow. There are a few key tests:
 - integration tests, run against an actual vCenter (Eco-vCenter)
 - sanity tests (includes a variety of linters)
-
```
anna-savina, line 37: It might be worth adding `Run tests locally` chapter, as vmware.vmware has

tests/integration/targets/vmware_rest_appliance-multi-hosts/tasks/main.yml
mikemorency: theres a bug with cloud.common that causes lookups to fail if environment is set as a variable. I opened a PR in their collection to fix it but they take a bit to merge and release

anna-savina (approved):

Issue #499, opened 2024-05-31
Allow disabling turbo mode


Alabate:
##### SUMMARY
We have an issue [with the use of the turbo server](https://github.com/ansible-collections/cloud.common/issues/158) in our project. This mean we cannot use the `vmware.vmware_rest` collection until we find a solution because there is no way to disable the turbo mode.

We could maybe add a condition on the `ENABLE_TURBO_MODE` env variable like in [kubernetes.core](https://github.com/ansible-collections/kubernetes.core/blob/0c5233a650c6c35d9621043fb107455bac3d9314/plugins/lookup/k8s.py#L189)

If you agree to this idea, I can create a PR.

##### ISSUE TYPE
- Feature Idea

PR #600, opened 2025-05-22
disable turbo mode in modules by default

mikemorency:
##### SUMMARY
This change will make the turbo server optional and disable it by default. This will slow down module execution but otherwise will not change the module behavior.

This is a user requested feature. However, with the recent ansible-core 2.19 changes it seems like now is a good time to implement this. We can add a statement saying the collection does not support turbo mode in 2.19+ and remove the statement if the upstream cloud.common collection works out all the issues.

Related issues: 
https://github.com/ansible-collections/vmware.vmware_rest/issues/499
https://github.com/ansible-collections/vmware.vmware_rest/issues/364

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/plugins/modules/appliance_access_consolecli.py
+++ b/plugins/modules/appliance_access_consolecli.py
@@ -109,7 +109,7 @@ from ansible.module_utils.basic import env_fallback
 import os
 
 try:
-    if not os.getenv('VMWARE_ENABLE_TURBO', False):
+    if not os.getenv("VMWARE_ENABLE_TURBO", False):
         raise ImportError()
 
     from ansible_collections.cloud.common.plugins.module_utils.turbo.exceptions import (
--- a/plugins/modules/appliance_access_consolecli.py
+++ b/plugins/modules/appliance_access_consolecli.py
@@ -108,20 +108,21 @@ PAYLOAD_FORMAT = {
 from ansible.module_utils.basic import env_fallback
 import os
 
-try:
-    if not os.getenv("VMWARE_ENABLE_TURBO", False):
-        raise ImportError()
+if os.getenv("VMWARE_ENABLE_TURBO", False):
+    try:
+        from ansible_collections.cloud.common.plugins.module_utils.turbo.exceptions import (
+            EmbeddedModuleFailure,
+        )
+        from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
+            AnsibleTurboModule as AnsibleModule,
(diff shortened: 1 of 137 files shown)


mikemorency:
@mariolenz  I am curious to hear what you think of this strategy, and if this would be considered a minor change or not

mikemorency:
Hm yea i do agree the `major_change` description sounds more accurate. I think it still would be a patch version release though, based on the [semver](https://semver.org/) descriptions:
```
MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
```

mariolenz:
Yes, it looks like a semver _major release_ and a changelog _major change_ differ significantly. 

The Ansible documentation about a major change says "include non-breaking changes". So I agree that a (semver) major release isn't necessary.

> Hm yea i do agree the `major_change` description sounds more accurate. I think it still would be a patch version release

To be precise: The current version is 4.7.0. I think 4.7.1 wouldn't be enough, it should be at least 4.8.0. But it wouldn't require 5.0.0. If this is what you meant, I'm with you there.

plugins/modules/appliance_access_consolecli.py
```diff
@@ -106,8 +106,12 @@
 }  # pylint: disable=line-too-long
 
 from ansible.module_utils.basic import env_fallback
+import os
 
 try:
+    if not os.getenv("VMWARE_ENABLE_TURBO", False):
+        raise ImportError()
+
```
mariolenz, line 114: I'm not sure about this. Looks a bit like misusing the ImportError exception here because, well, we don't have a problem importing something. Or do we? This doesn't look too great to me.

Although this means some additional lines, I think this would be better:

```
if os.getenv("VMWARE_ENABLE_TURBO", True):
    try:
        ...
    except ImportError:
        from ansible.module_utils.basic import AnsibleModule
else:
    from ansible.module_utils.basic import AnsibleModule
```

This would make the logic clearer.

It would be great to hear the opinion of other people about this! After all, I might be wrong there.

mariolenz (approved):
I didn't check _all_ 137 changed files, but what I've seen LGTM!

It's a little bit repetitive to have this if / try / except / else block in all the files, but I don't know a better solution right now. At the end of the day, this is not a new issue. It's already been repetitive to have this try / except everywhere, now it's only an additional if / else.

So let's keep it as-is and maybe implement a better solution later.

PR #608, opened 2025-06-09
Feature/deprecate lookups

mikemorency:
##### SUMMARY
This deprecates the lookup plugins in favor of vmware.vmware.moid_from_path, since the lookup plugins in this repo will not work with ansible-core 2.19 and above

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vmware.vmware_rest.*_moid


mariolenz:
It looks like the docs test fails because the moid_from_path lookup plugin isn't part of a release yet. But I'm not 100% sure there.

I don't understand why the integration tests fail... could this be something temporarily?

p3ck (approved):
other than the type of changelog fragment it looks good to me

changelogs/fragments/608-deprecate-lookup-plugins.yml
```diff
@@ -0,0 +1,4 @@
+---
+trivial:
```
mariolenz, line 2:
```suggestion
deprecated_features:
```

I think @p3ck is right, this should be `deprecated_features` and not `trivial`.

Issue #617, opened 2025-07-23
vcenter_vm_hardware_ethernet module performs actions for real in check mode


ryanb74:
### Summary

This module is described as supporting check mode however there is no logic in the python code to handle this, it will run for real whenever ansible is running in check mode or not !

A quick fix would be to change the flag "support_check_mode" to false in that module code to avoid people having unexpected changes made

### Issue Type

Bug Report

### Component Name

vcenter_vm_hardware_ethernet module

### Ansible Version

```console (paste below)
$ ansible --version
2.18.5
```

### Collection Versions

```console (paste below)
$ ansible-galaxy collection list
vmware.vmware_rest      4.7.0
```

### Steps to Reproduce

Run the module in check mode to add or remove a network interface

### Expected Results

The module should report "changed" if it WOULD do changes, without actually doing anything

### Actual Results

The module reports "changed" and DOES changes, even though check mode is activated 

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

ryanb74:
> duplicate of [#363](https://github.com/ansible-collections/vmware.vmware_rest/issues/363)
> 
> This is a known issue, and a limitation of how this collections content is generated

Could you be more speific and explain what's going on exactly with this "collection content generation" ?

It doesn't seem impossible to set the support check mode flag to false like I did in this merge request : 
https://github.com/ansible-collections/vmware.vmware_rest/pull/618/files

mikemorency:
The modules in this collection are generated based on templates in a different collection. We get a REST API spec from broadcom, feed that into the generator, and get modules back. When a change is made directly to the modules, it is overwritten the next time modules are generated. You can read more about it here: https://github.com/ansible-collections/vmware.vmware_rest/blob/main/docs/development.md#how-to-generate-collection-modules

Local changes can be made, but unless we can make additional changes upstream, the problem is not resolved

PR #618, opened 2025-07-23
fix(vcenter_vm_hardware_ethernet): set `supports_check_mode` flag to …

ryanb74:
…False

This module doesn't implement check mode support and runs for real even in check mode, hence this change

##### SUMMARY
Check the following issue :
https://github.com/ansible-collections/vmware.vmware_rest/issues/617

### DIFF
--- a/plugins/modules/vcenter_vm_hardware_ethernet.py
+++ b/plugins/modules/vcenter_vm_hardware_ethernet.py
@@ -436,7 +436,7 @@ async def main():
 
     module_args = prepare_argument_spec()
     module = AnsibleModule(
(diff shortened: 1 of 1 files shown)


mikemorency:
None of the modules actually support check mode, but they all say they do. We should change it upstream, but I still have a [PR](https://github.com/ansible-community/ansible.content_builder/pull/89) from a year ago attempting to change something that they have not merged

I think given people's issues and the upstream lack of activity, it makes sense to change the setting in the existing modules and just update check mode to false in my existing PR

mikemorency:
Closing this. Superseded by https://github.com/ansible-collections/vmware.vmware_rest/pull/626

PR #620, opened 2025-08-22
remove cloud.common from lookup plugins

mikemorency:
##### SUMMARY
Removing the cloud.common dependency from lookup plugins. This will impact performance negatively

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
lookup plugins

### DIFF
--- a/plugins/lookup/cluster_moid.py
+++ b/plugins/lookup/cluster_moid.py
@@ -129,10 +129,9 @@ _raw:
 """
 
 
-from ansible_collections.cloud.common.plugins.plugin_utils.turbo.lookup import (
-    TurboLookupBase as LookupBase,
-)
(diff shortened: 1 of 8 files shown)


mikemorency:
@mariolenz I have been trying different async loop stuff without success. It seems like the issue happens if I call the plugins more than once in the same playbook. The async loop keeps getting closed in the process, and then I cant create a new one for some reason.

I think it might be possible, but this is not a typical use case for asyncio, and I am not an expert with that library. I do not think I can get this working anytime soon. I believe we should remove the cloud.common dep and print a helpful error for users

mariolenz:
> I do not think I can get this working anytime soon. I believe we should remove the cloud.common dep and print a helpful error for users

Let me work a bit on a suggestion.

PR #621, opened 2025-08-26
Drop cloud.common requirement

### LINKED ISSUE #619: Drop dependency on cloud.common
Following up the discussion in ansible-collections/cloud.common#166, since I don't want to clutter it. Let's move this to here.

I'd like to keep this collection in the Ansible Community Package if possible, at least for now.

As far as I remember, #600 made `cloud.common` more or less optional. That is, it's only used when `VMWARE_ENABLE_TURBO` is `True`. So the collection doesn't really need `cloud.common`, it's only needed when people want to use this turbo thing. This looks like we could remove the dependency.

BUT I think the lookup plugins, or at least some, always import `cloud.common`. We should make this optional there too. Then I think we could remove `cloud.common` as a dependency.

What do you think @mikemorency?

CC @felixfontein

mariolenz:
Fixes #619

##### SUMMARY
Since [cloud.common isn't compatible with ansible-core 2.19](https://forum.ansible.com/t/41507/24), I think we should drop it as a hard dependency. That is: From `galaxy.yml`.

It would be better if the lookup plugins would work without or make this turbo stuff only optional there. But since this doesn't look easy, my solution is to 1) mention in the documentation that this collection is needed and 2) fail if it isn't installed.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
cluster_moid
datacenter_moid
datastore_moid
folder_moid
host_moid
network_moid
resource_pool_moid
vm_moid

##### ADDITIONAL INFORMATION
#620
ansible-collections/cloud.common#183

### DIFF
--- a/changelogs/fragments/619-drop-cloud.common-requirement.yml
+++ b/changelogs/fragments/619-drop-cloud.common-requirement.yml
@@ -1,4 +1,4 @@
 ---
-minor_changes:
+major_changes:
   - Remove ``cloud.common`` as a dependency, so it will not be installed automatically anymore
     (https://github.com/ansible-collections/vmware.vmware_rest/pull/621).
--- a/changelogs/fragments/619-drop-cloud.common-requirement.yml
+++ b/changelogs/fragments/619-drop-cloud.common-requirement.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Remove ``cloud.common`` as a dependency, so it will not be installed automatically anymore
+    (https://github.com/ansible-collections/vmware.vmware_rest/pull/621).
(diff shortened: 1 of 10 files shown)


mikemorency (approved):

mikemorency:
Seems odd to me that the tests don't fail. Cloud.common must still be present on the test host. I can look more into it, but this should still be good to merge in the meantime

mariolenz:
@felixfontein Do you see any problems doing this in a minor release?

felixfontein:
I think it's borderline, but I think it should be OK - also since Galaxy as a package manager isn't that flexible, since you can't say that cloud.google is only needed for ansible-core < 2.19 or something like that, or that it is an optional/recommended/... dependency.

I'd probably change the changelog fragment from `minor_changes` to `major_changes` though, so that it both shows up in the Ansible porting guide and that it's clearer for users that this is a potentially more important change.

mariolenz:
> I think it's borderline, but I think it should be OK

So many things feel borderline ATM :weary: 

> I'd probably change the changelog fragment from `minor_changes` to `major_changes` though

Done. And I've also added a known issue.

@mikemorency If this looks OK to you, do you think you can merge and release 4.9.0?

PR #627, opened 2025-12-05
deprecate modules moved to vmware.vmware

mikemorency:
##### SUMMARY
Deprecating modules that are made redundant by the vm and vm_apply_customization modules in vmware.vmware

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
    vcenter_vm_guest_customization,
    vcenter_vm_hardware_adapter_sata,
    vcenter_vm_hardware_adapter_scsi,
    vcenter_vm_hardware_cdrom,
    vcenter_vm_hardware_cpu,
    vcenter_vm_hardware_disk,
    vcenter_vm_hardware_ethernet,
    vcenter_vm_hardware_memory,
    vcenter_vm

### DIFF
--- a/changelogs/fragments/625-deprecate-modules.yml
+++ b/changelogs/fragments/625-deprecate-modules.yml
@@ -1,5 +1,5 @@
 ---
-minor_changes:
+deprecated_features:
   - Deprecate modules that have been moved to the new vmware.vmware collection. Includes
     vcenter_vm_guest_customization,
     vcenter_vm_hardware_adapter_sata,


mariolenz:
The failing Build Ansible Docs job doesn't look great to me. This might be fixed by adding the vmware.vmware modules to provide-link-targets in [.github/workflows/docs-pr.yml](https://github.com/ansible-collections/vmware.vmware_rest/blob/main/.github/workflows/docs-pr.yml) like in #551.

On the other hand, provide-link-targets seems to fake valid references / labels / links. So if there's a typo in the one place and we "fix" this by copying this over over to provide-link-targets, there's still a problem.

Maybe installing vmware.vmware in [extra-collections](https://github.com/ansible-community/github-docs-build/blob/00a4f624b315edca02a6fcf7ab306b39f932a36e/.github/workflows/_shared-docs-build-pr.yml#L161-L164) might be a solution?

I think I'll give this a try.

mariolenz:
I've added ansible_collections.vmware.vmware.vm_apply_customization_module and ansible_collections.vmware.vmware.vm_module to Validate Ansible Docs. This looks better. I still don't understand why we need this here but not in Build Ansible Docs :shrug:

mikemorency:
Whatever the issue was, it went away. I think the tests can be updated/removed to use different modules in a few places. Ill open another PR after this one to do that

PR #628, opened 2025-12-06
Build Ansible Docs: extra-collections instead of provide-link-targets

mariolenz:
##### SUMMARY
This collection refers to modules from another collection: vmware.vmware. This fails the Build Ansible Docs CI job because it doesn't know about it.

It seems that the current fix is to fake the missing references with provide-link-targets. I wonder if it wouldn't be better to make the job aware of vmware.vmware via extra-collections instead.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
.github/workflows/docs-pr.yml

##### ADDITIONAL INFORMATION
#627

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -23,6 +23,18 @@ jobs:
       artifact-name: ${{ github.event.repository.name }}_docs_${{ github.event.pull_request.head.sha }}_validate
       intersphinx-links: |
         ansible_devel:https://docs.ansible.com/ansible-core/devel/
+      provide-link-targets: |
+        ansible_collections.vmware.vmware.appliance_info_module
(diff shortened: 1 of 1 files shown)


mariolenz:
In the first commit, I've simply removed provide-link-targets. Just wanted to see what happens. There are a lot of [undefined label warnings](https://github.com/ansible-collections/vmware.vmware_rest/actions/runs/19988784062/job/57326497503?pr=628) as expected.

Now let's try to add extra-collections.

mariolenz:
@mikemorency I wonder why extra-collections works for build-docs, but not for validate-docs... anyway, it still might be an improvement.

mikemorency (approved):

Issue #637, opened 2026-02-19
vcenter_host_info references cloud.common that doesn't support Ansible 2.20


MallocArray:
### Summary

When a task runs module `vmware.vmware_rest.vcenter_host_info` with Ansible 2.20, a warning appears in the output, while the task completes successfully.

`[WARNING]: Collection cloud.common does not support Ansible version 2.20.2`

Environment variables are not set to enable turbo mode

### Issue Type

Bug Report

### Component Name

vcenter_host_info

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.20.2]
  config file = /runner/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.12/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections:/runner/collections
  executable location = /usr/local/bin/ansible
  python version = 3.12.12 (main, Jan 16 2026, 00:00:00) [GCC 14.3.1 20250617 (Red Hat 14.3.1-2)] (/usr/bin/python3.12)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

### Collection Versions

```console (paste below)
$ ansible-galaxy collection list
vmware.vmware_rest        4.10.0 
cloud.common              5.0.0 
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
COLLECTIONS_PATHS(/runner/ansible.cfg) = ['/root/.ansible/collections', '/usr/share/ansible/collections', '/runner/collections']
CONFIG_FILE() = /runner/ansible.cfg
DEFAULT_FILTER_PLUGIN_PATH(/runner/ansible.cfg) = ['/runner/collections/ansible_collections/mts/common/plugins/filter']
DEFAULT_ROLES_PATH(/runner/ansible.cfg) = ['/runner/roles']
DEFAULT_STDOUT_CALLBACK(/runner/ansible.cfg) = community.general.timestamp
DEFAULT_TIMEOUT(/runner/ansible.cfg) = 40
HOST_KEY_CHECKING(/runner/ansible.cfg) = False
INTERPRETER_PYTHON(/runner/ansible.cfg) = auto_silent
```

### OS / Environment

Execution Environment build on UBI10

### Steps to Reproduce

```yaml (paste below)
- name: Gather hosts in cluster
[...]

PR #639, opened 2026-05-05
Deprecate VMWARE_ENABLE_TURBO

mariolenz:
##### SUMMARY
`cloud.common` doesn't support ansible-core 2.20, and it doesn't look like anyone is working on this. Actually, not supporting current ansible-core versions has been the reason to remove it from the Ansible Community Package (ansible-community/ansible-build-data#593).

We've made `cloud.common` optional in #600 but I think it's time to get rid of this collection completely by now. But in order to do this, we have to deprecate turbo mode / VMWARE_ENABLE_TURBO because enabling this also means using `cloud.common`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Lots, probably all.

##### ADDITIONAL INFORMATION
#637

### DIFF
--- a/docs/turbo_mode.md
+++ b/docs/turbo_mode.md
@@ -1,5 +1,7 @@
 # vmware.vmware_rest Turbo Mode
 
+!!! Please be aware that this is deprecated and will be removed in a future major release. !!!
+
 This collection is capable of leveraging a feature of the cloud.common collection called the "turbo server". This is basically a caching mechanism that speeds up repeated API calls. This collection has used turbo mode up until version 4.8.0. With the release of 4.8.0, turbo mode is disabled by default but can be re-enabled using an environment variable.
 
 Read more about the turbo server in the cloud.common documentation [here](https://github.com/ansible-collections/cloud.common?tab=readme-ov-file#ansible-turbo-module).
--- a/docs/turbo_mode.md
+++ b/docs/turbo_mode.md
@@ -1,6 +1,7 @@
-# vmware.vmware_rest Turbo Mode
+> [!WARNING]
+> Please be aware that Turbo Mode is deprecated and will be removed in a future major release.
 
-!!! Please be aware that this is deprecated and will be removed in a future major release. !!!
+# vmware.vmware_rest Turbo Mode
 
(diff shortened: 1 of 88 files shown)


mariolenz:
@mikemorency What do you think about this? It'll be a lot of copy and paste to add the deprecation warning to all modules, but I don't have any better idea just now.

docs/turbo_mode.md
```diff
@@ -1,5 +1,7 @@
 # vmware.vmware_rest Turbo Mode
 
+!!! Please be aware that this is deprecated and will be removed in a future major release. !!!
```
mikemorency: maybe formatting like this might be better, https://github.com/orgs/community/discussions/16925

mariolenz: Good point! But how should we document this? As important, warning or caution?

I tend to either warning or caution, but I'm not sure what would be better. @mikemorency Do you have a preference there?

mikemorency: No preference really, but I agree with warning or caution. Purple just seems weird 🤔

mikemorency:
I think it is good. Should we also add a note to the module docs about the deprecation or cloud.common warning? The turbo docs are linked in the main README [here](https://github.com/ansible-collections/vmware.vmware_rest/blob/4.10.0/README.md#slower-execution-times), but as you pointed out there is no path from the module docs (which are the main thing people read) explaining the cloud.common warning

mariolenz:
As VMWARE_ENABLE_TURBO isn't mentioned in the modules documentation right now, I think we shouldn't invest the time to ducument that a) the environment variable exists and b) it wouldn't be a good idea to start using it because it will be removed in 5.0.0.

Talking about being lazy, I've added the warning only to those modules that are not scheduled to be removed in 5.0.0. It's been less work, and for users this information is irrelevant because in 5.0.0 they won't be able to use the module at all, neither with nor without VMWARE_ENABLE_TURBO.

BTW I don't have access to my usual test environment at the moment. Maybe you could run a test or two to make sure that my changes really work.

mikemorency (approved):

mariolenz:
Thanks for reviewing @mikemorency!

Issue #609, opened 2025-06-25
Use official OpenAPI specification


mariolenz:
### Summary

Idea: Use the official [vSphere OpenAPI Specification](https://developer.broadcom.com/sdks/vcf-api-specification/latest/) to build the collection.

### Issue Type

Feature Idea

### Component Name

Well, everything :-)

### Additional Information

When I tried to get an API specification for 7.0 U3, this has been a real pain: #449

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

mikemorency:
Im going to close this since the newly generated modules are using this spec

PR #646, opened 2026-06-09
add support for 2.21

mikemorency:
##### SUMMARY
Add testing support for ansible 2.21

Also update the minimum version in the runtime.yml to 2.16. This is required for publishing a new version to automation hub.

##### ISSUE TYPE
- Feature Pull Request


.github/workflows/sanity-tests.yaml
mariolenz: I think you have to remove stable-2.15 here, otherwise the sanity tests fail with:

> Collection vmware.vmware_rest does not support Ansible version

mariolenz (changes requested):

changelogs/fragments/09062026-add-2.21.yml
```diff
@@ -0,0 +1,6 @@
+---
+minor_changes:
+  - Add support for ansible-core 2.21
+
+major_changes:
+  - Update minimum required ansible-core version to 2.16 in meta/runtime.yml
```
mariolenz, line 6: I'm not really sure about this. You see, dropping support for an ansible-core version might be considered a breaking instead of a major change.

I don't know if we really have reached a consensus about this in the community. But if it's OK for you, I would make this a breaking change and release 5.0.0 instead of a major change and do another 4.x release.

mariolenz (commented):
I'm not really sure about this. You see, dropping support for an ansible-core version might be considered a breaking instead of a major change.

I don't know if we really have reached a consensus about this in the community. But if it's OK for you, I would make this a breaking change and release 5.0.0 instead of a major change and do another 4.x release.

mariolenz:
BTW a new major release would also allow us to get rid of some deprecated stuff, which might make it a little bit easier to maintain the collection: https://github.com/ansible-collections/vmware.vmware_rest/milestone/1

Although we might want to ignore #609 and move it to 6.0.0.

changelogs/fragments/09062026-add-2.21.yml
mikemorency: Additional context: I cannot publish this collection to automation hub until the minimum version is updated.

So my train of thought is that its been awhile since we have done a minor release, and there are probably some nice to have changes that people would appreciate having. If we do a major release, it will be harder for them to get those nice to haves without also dealing with the many deprecation we have pending.
So id like to do this as part of a minor release, and then cut a major release pretty much right after.

And the only reason im considering this is because of the blocker to automation hub uploads, and the fact that 2.15 is so old. WDYT?

Yaish25491 (approved):
lgtm

changelogs/fragments/09062026-add-2.21.yml
mariolenz: As I've said, IIRC there's not really a consensus in the community about dropping older versions of ansible-core or other requirements. I would try to make this a breaking change and do a new major release if possible, but since ansible-core 2.15 reached EOL Nov 2024... :shrug:

mariolenz (approved):
LGTM

mariolenz:
@mikemorency Since you think this a good idea, @Yaish25491 approved and I changed my mind about the major / breaking change issue I've merged.

I hope that's OK for you.

PR #650, opened 2026-06-16
remove old api specs

mikemorency:
##### SUMMARY
Removing old and unsupported api specs

##### ISSUE TYPE
- Docs Pull Request


mariolenz (approved):
I don't think it hurts to keep those specs in the repo. On the other hand it doesn't hurt to remove them, either.

Since vSphere 6 and 7 are EOL, let's remove them. We don't need those specs anymore for anything in this collection. They're still available in the git history somewhere if someone really wants or needs to them.

PR #651, opened 2026-06-17
[WIP] Regenerate modules using agentic workflow

mikemorency:
##### SUMMARY
This PR regenerates all modules using an agentic workflow. The agent creates the modules based on a vSphere API spec. This includes
- unit tests for all modules
- integration tests for all modules, using a mock api server
- documented support for vSphere API 7-9
- workflow updates to run the new tests

##### ISSUE TYPE
- Feature Pull Request


mariolenz:
587 changed files are _a lot_ to review. I don't think I can do it. You've generated all this using AI, and someone can "review" by asking an AI if it's OK. But then one AI would check another. Call me old-fashioned, but I like to have a human in the reviewing loop. Someone who is really reviewing and not just asking an AI and nodding through what it says. But reviewing 587 files is just too much...

grafik

bcoca:
from quick scan:
- all lookup plugins deleted, no reason, no deprecation  
- many required files deleted, including requirements.txt 
- much of the metadata removed, including current deprecations

probably a few hundred other issues, but as per above, not sure how this will ever be fully reviewed

mikemorency:
@mariolenz I both appreciate and share some skepticism about AI generated content. If you have any suggestions on validation, please let me know. Ill be sure to include them.

I think there is a lot of potential in the idea of generating content based on an API spec/documentation, with some minimal scaffolding. For example, these changes include a massive increase in both test and documentation coverage. Its not a solution that works for all, or even most, collections. But this is not a typical collection. Given that this content has always been "generated" based on the API, I feel that the biggest difference between this round of generation and the round for v4.X is that we're aware of the things we dont really have confidence in.

Im not 100% convinced that this will all work out. I havnt even begun to run these modules in the old integration test suite. But I do feel like this is worth exploring, and there are few collections out there in a better position to do so. Testing and validation are challenges, but they have always been a challenge in this collection, especially when things were first generated.

bcoca:
In my experience, generated modules have always been of much lower quality than their peers. Most APIs are not very user friendly nor conducive to good UX.

Consider that many Ansible users are not programmers, one of the main features was making it accessible and intelligible to a less tech savvy user base.  I'm not against generating the module code using the API as a base, but I'm very much wary of that being just a straight pass through to the user's input/interface.

As for 'reviewability' i would suggest starting small, do one module/plugin first, once the process is perfected you can move to adding the LLM helper files and continue adding a feature at a time. 

LLMs can do things at scale that humans cannot, but LLMs lack 'common sense', have a very limited view of context and history and are designed to give a well structured response more than giving a correct response.  They are great at volume, but not as great at quality. This is why I still consider that human auditability should be the primary concern.

mikemorency:
@bcoca 
> In my experience, generated modules have always been of much lower quality than their peers. Most APIs are not very user friendly nor conducive to good UX.

That has definitely been the experience I have had with this collection. Previous iterations of the modules have had missing or nonsensical documentation, partial functionality, and poor maintainibility. In my limited experience, the agentic workflow has made those issues better. Like I said in my previous comment, I don't think this is the best approach for most collections. Generated content (ai or not) is a niche solution that has real drawbacks. In reality though, this content would not exist if it was not generated, and people do use it.

As for reviewing, I'd welcome anyone's constructive review on any module(s). I don't expect a full review on all modules, and that has never been expected when generating content in this collection.

I didn't just open this draft on a whim. I iterated with one module, then 5, then 10, then 30.... Until I got to a point where the entire collection could be generated with consistent results. Then I opened the draft to leverage CI for validation. And after this, I have plans for regression tests. Even after all of that, I do plan on asking other maintainers to take a focused look at more than one, but not all, modules in a new and polished pr. 

I agree AI generated content deserves skepticism and a thorough review from actual developers. I am also wary of an initial feeling of being overwhelmed, and potentially dismissing an improvement because it presents new challenges.

mariolenz:
> I both appreciate and share some skepticism about AI generated content. If you have any suggestions on validation, please let me know. Ill be sure to include them.

@mikemorency You don't have to listen to me, I think I'm a _little_ bit biased when it comes to AI :smirk: 

I haven't seen any real benefit from AI for me yet. On the other hand, it uses a lot of energy which makes it even harder to fight climate change than it already is. And it led to make my hardware far too expansive :rage: 

Well, I don't want to be an old man yelling at clouds. And maybe it's about time that I have a closer look at where AI can be helpful. I don't think it will go away.

So I try to suggest something that is hopefully helpful instead of just ranting around: How about splitting creating the modules and the integration tests? Using AI to generate all the modules and all the integration tests in one go feels weird to me. How about using AI to generate the modules and then check if the CI still works? Or do it the other way round and generate new integration tests with AI and see what happens. Then compare the result. This way, we could decide if the AI generated stuff or the old / manually created stuff is wrong or buggy.

> In my experience, generated modules have always been of much lower quality than their peers. Most APIs are not very user friendly nor conducive to good UX.
> 
> Consider that many Ansible users are not programmers, one of the main features was making it accessible and intelligible to a less tech savvy user base. I'm not against generating the module code using the API as a base, but I'm very much wary of that being just a straight pass through to the user's input/interface.

@bcoca I think you're right. APIs follow a different logic. They're OK for programmers, but modules generated from an API specification usually aren't very user-friendly. They work like the API does, not like end users think about doing things to reach a definite result.

> As for 'reviewability' i would suggest starting small, do one module/plugin first, once the process is perfected you can move to adding the LLM helper files and continue adding a feature at a time.

I think that's a sound advice. Instead of regenerating practically everything in one go, let's start somewhat smaller. This won't fix the basic problem that modules generated from an API spec usually aren't very user-friendly, though.

mikemorency:
@mariolenz I am not sure about breaking up the test and module generation. They work in tandem, forcing the agents to iterate and fix issues with the module. Ill explain in more detail below.

I can re-introduce the old integration tests and have them continue to run though. That way we can see the regression results in each PR. Ill also try to add some more manual integration tests. Since the AI generated ones cover things like idempotence, check_mode, and edge cases it should be easier to write some new tests to cover the basics. Once we are ready to merge (or release 5.0), I will want to remove the old tests. But at that point we should have confidence in the new tests.

I can also split this into smaller groups. It makes no difference to the AI if they are all generated, or just a subset, so its no problem. I could see that give us more confidence in the tests' ability to detect issues.

Let me know what you think. 

RE test generation:
There are multiple subagents, each with rules to not touch files outside of their domain.
One agent reads the API spec and uses it to generate a module, using a template with a base class interface and module utils (that i wrote). Theres also another rule for this agent to not look at other modules in the collection so the templates/references are always used and bad feedback loops are avoided.
Then, a second agent generates unit tests, using the API spec as the source of truth for expected results, inputs, and API calls. If failures are found, the test agent sends the results to the module agent and asks it to fix them in the module only.
Once unit tests pass, a similar iteration is done for integration tests. The integration subagent uses the module documentation as the source of truth, and mocks the API using the spec.
There are additional rounds of subagents for documentation consistency, formatting, sanity/lint. And after each round, if functional changes are done to the module, tests are run again to ensure no breaks.
(And I do note that all of this is assumes the AI does what it says it will do, of course :slightly_smiling_face: )

PR #652, opened 2026-06-18
remove miscellaneous files that are not used by anything

mikemorency:
##### SUMMARY
Removing some files that are not used by anything

##### ISSUE TYPE
- Chore Pull Request


bcoca:
These files are used, just not by the code in this repo, consider that other tools like ansible-builder and ansible-test need these files to properly interact with the collection.

mariolenz (approved):
It looks like those files really aren't used anywhere, so LGTM.

If it turn's out that those files are important for anything, it will be easy to restore them. So let's merge.

PR #648, opened 2026-06-15
tombstone deprecated plugins

mikemorency:
##### SUMMARY
Tombstone all the modules that were previously deprecated, in preparation for 5.0.0
Fixes
https://github.com/ansible-collections/vmware.vmware_rest/issues/629
https://github.com/ansible-collections/vmware.vmware_rest/issues/630
https://github.com/ansible-collections/vmware.vmware_rest/issues/631

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
There are many

PR #654, opened 2026-06-20
[WIP] Fix integration tests after removing modules

mariolenz:
##### SUMMARY
Fixing the integration tests that have been broken by removing deprecated modules in #648.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
lots


mikemorency:
FWIW I was looking at these yesterday and a lot of the test content is no longer relevant, since modules were deprecated. I think at least one test case was removed entirely

mariolenz:
> Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

I've never seen [that](https://github.com/ansible-collections/vmware.vmware_rest/actions/runs/27870178351/job/82480716976?pr=654) before. @mikemorency Do you know if this is something new in actions/checkout@v7 (#653)?

This is a little bit awkward because I'm used to prepare my PRs in a fork. Often, I don't even have permissions to do anything in the real repo itself.

mikemorency:
It does look related to that new version of the checkout action. We have the manual approval gate in place for this integration test though so we should be able to safely use the option the error mentions to get it working again

mariolenz:
> We have the manual approval gate in place for this integration test though so we should be able to safely use the option the error mentions to get it working again

Should I create a PR to add `allow-unsafe-pr-checkout: true`?

mikemorency:
sorry, i didnt see the follow up message. I opened a pr to fix: https://github.com/ansible-collections/vmware.vmware_rest/pull/655

mikemorency:
I made some progress: https://github.com/ansible-collections/vmware.vmware_rest/pull/656
I have everything working except the vm_scenario test. It just needs a closer look

mikemorency:
@mariolenz i think this can be closed now

mariolenz:
I guess you're right.

PR #656, opened 2026-06-20
update integration tests

mikemorency:
##### SUMMARY
Fixing integration tests by removing or replacing tombstoned modules/tests

I also restructured the VM tests a bit, since there was a lot of unrelated or redundant files/tests 

##### ISSUE TYPE
- Bugfix Pull Request


tests/integration/targets/vmware_rest_library_and_ovf_clone/tasks/main.yml
```diff
@@ -34,83 +34,35 @@
...
         source:
           type: VirtualMachine
           id: "{{ test_vm.vm.moid }}"
         target:
-          library_id: "{{ nfs_lib.id }}"
+          library_id: "{{ ci_resources_content_library_id }}"
         create_spec:
-          name: "{{ ovf_image_name }}"
-          description: an OVF example
+          name: "{{ vm_from_ovf_image_name }}"
```
p3ck, line 64: Where is vm_from_ovf_image_name defined?  I didn't see it when I expanded but I'm sure I just missed some other file where it is?

p3ck (approved):
Just one question on where that variable is defined.

tests/integration/targets/vmware_rest_library_and_ovf_clone/tasks/main.yml
mikemorency: its in the role's defaults

PR #663, opened 2026-07-09
regenerate datacenter modules and add tests

mikemorency:
##### SUMMARY
Regenerating the vcenter_datacenter* modules using the new scripts and AI.

Also includes generated unit tests and integration tests.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vcenter_datacenter
vcenter_datacenter_info


mikemorency:
@mariolenz please review and let me know what you think. these are the first modules I have generated

Yaish25491:
```markdown
# PR #663 — `vcenter_datacenter` Module Regeneration — Code Review

**Repo:** ansible-collections/vmware.vmware_rest  
**PR:** [#663](https://github.com/ansible-collections/vmware.vmware_rest/pull/663) — "regenerate datacenter modules and add tests"  
**Review effort:** High (8 finder angles, 6 verified)  
**Result:** 7 findings, all CONFIRMED. 0 refuted.

---

## Finding 1 — `force` parameter silently dropped on DELETE

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter.py` |
| **Location** | `DELETE_OPERATION` definition |
| **Severity** | Critical |
| **Verdict** | CONFIRMED |

**Summary:**  
`DELETE_OPERATION` has no `query_spec`, so the `force` parameter is accepted by the argument spec but never transmitted as `?force=true` on the DELETE request. The old module correctly mapped `force` via `PAYLOAD_FORMAT["delete"]["query"] = {"force": "force"}`.

**Failure scenario:**  
User runs `force: true` to delete a non-empty datacenter. API returns `ResourceInUse` error because `?force=true` was never appended.

**Fix:**  
Add `query_spec={"force": {"required": False}}` to `DELETE_OPERATION`.

**Test gap:**  
`test_delete_with_force` only asserts `delete.assert_called_once()`, never checks that `force` reached the query params. `test_delete_datacenter` explicitly asserts `query is None`, codifying the bug.

---

## Finding 2 — Folder-based lookup lost in idempotency check

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter.py` |
| **Location** | `LIST_OPERATION` + `_search_for_resource` |
| **Severity** | Critical |
| **Verdict** | CONFIRMED |

**Summary:**  
The old module used `lookup_with_filters` to query `?names={name}&folders={folder}`. The new `LIST_OPERATION` has no `query_spec`, so `_search_for_resource` fetches ALL datacenters and matches by name only, ignoring `folder`. The info module's `LIST_OPERATION` correctly includes `query_spec` with `names`, `folders`, `datacenters` — the CRUD module simply omitted them.

**Failure scenario:**  
Two datacenters named `prod-dc` in different folders. Module matches the first one found, may skip intended creation or operate on the wrong datacenter.

**Fix:**  
Add `query_spec={"names": {"required": False}, "folders": {"required": False}}` to `LIST_OPERATION`.

---

## Finding 3 — Return value structure breaking change

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter_info.py` |
| **Location** | `RETURN` block |
| **Severity** | High (breaking change) |
| **Verdict** | CONFIRMED |

**Summary:**  
Old module returned `value` as `type: list` containing all results. New module returns `value` as `type: dict` (single resource only) and adds `info` as `type: list`. No backward-compatibility shim exists.

**Failure scenario:**  
Existing playbooks using `result.value | length` or `with_items: '{{ existing_datacenters.value }}'` (shown in old EXAMPLES) will fail. `value` is now either undefined (list mode) or a dict (single-resource mode), never a list.

---

## Finding 4 — Parameter aliases dropped without deprecation

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter_info.py` |
| **Location** | `create_module_argument_spec()` |
| **Severity** | High (breaking change) |
| **Verdict** | CONFIRMED |

**Summary:**  
The old module defined `aliases: [filter_datacenters]` for `datacenters`, `aliases: [filter_folders]` for `folders`, and `aliases: [filter_names]` for `names`. All three aliases are removed with no `deprecated_aliases` mechanism or migration path.

**Failure scenario:**  
Playbooks using `filter_datacenters`, `filter_folders`, or `filter_names` fail with "Unsupported parameters" after upgrade.

---

## Finding 5 — `version_added` incorrectly changed from 0.1.0 to 5.0.0

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter.py`, `vcenter_datacenter_info.py` |
| **Location** | `DOCUMENTATION` block |
| **Severity** | Medium |
| **Verdict** | CONFIRMED |

**Summary:**  
Both modules have existed since `0.1.0`. The PR changes `version_added` to `5.0.0`, falsely presenting them as brand new modules. `version_added` reflects when a module was first introduced, not when it was rewritten.

**Failure scenario:**  
`ansible-doc` and Galaxy show these as "new in 5.0.0", confusing users who have used them since `0.1.0`. Porting guides may incorrectly list them as new additions.

---

## Finding 6 — `folder` not enforced as required for create

| Field | Value |
|-------|-------|
| **File** | `plugins/modules/vcenter_datacenter.py` |
| **Location** | argument spec + `CREATE_OPERATION` |
| **Severity** | Medium |
| **Verdict** | CONFIRMED (pre-existing, carried forward) |

**Summary:**  
Documentation states "This is required when creating a datacenter" but `CREATE_OPERATION.body_spec` marks `folder` as `required: False` and there is no `required_if=["state", "present", ["folder"]]`. The unit test `test_create_datacenter_name_only` explicitly tests without `folder` and expects success, contradicting the docs.

**Failure scenario:**  
User omits `folder` on create, passes all Ansible validation, gets opaque vSphere API 400 error instead of clear validation message.

**Note:** The old module had the same gap. This is a pre-existing issue, not a regression.

---

## Finding 7 — Unit test codifies the `force` bug

| Field | Value |
|-------|-------|
| **File** | `tests/unit/plugins/modules/test_vcenter_datacenter.py` |
| **Location** | `test_delete_with_force`, `test_delete_datacenter` |
| **Severity** | Medium (test quality) |
| **Verdict** | CONFIRMED |

**Summary:**  
`test_delete_with_force` passes `force: True` but only asserts `changed is True` and `delete.assert_called_once()`. It never verifies that `force=True` appears in the DELETE request's query parameters. Meanwhile, `test_delete_datacenter` asserts `call_args[1]["query"] is None`, actively locking in the broken behavior.

**Fix:**  
Add `assert call_args[1]["query"] == {"force": True}` to the force test.

---

## Summary Table

| # | Finding | Severity | Verdict |
|---|---------|----------|---------|
| 1 | `force` silently dropped on DELETE | Critical | CONFIRMED |
| 2 | Folder-based lookup lost (idempotency broken) | Critical | CONFIRMED |
| 3 | `value` return type changed from list to dict | High (breaking) | CONFIRMED |
| 4 | Parameter aliases dropped without deprecation | High (breaking) | CONFIRMED |
| 5 | `version_added` incorrectly bumped to 5.0.0 | Medium | CONFIRMED |
| 6 | `folder` not enforced as required (pre-existing) | Medium | CONFIRMED |
| 7 | Test codifies the `force` bug | Medium (test) | CONFIRMED |

## Recommended Priority

1. Fix `DELETE_OPERATION` to include `query_spec` for `force`
2. Fix `LIST_OPERATION` to include `query_spec` for `names`/`folders`
3. Decide on backward-compat strategy for `value` return type and parameter aliases
4. Restore `version_added: 0.1.0`
```

Yaish25491 (approved):

mikemorency:
I am fixing 1, 2, 4, 7
3 - The old modules have inconsistent and incorrectly documented return values. I did my best to maintain backwards compatibility but it may not be 100%. The new returns are documented and consistent, so at least it will be fixed going forward
5 - I think this is actually a good thing, since we should regenerate less often now and this regeneration is significant
6 - This is all correct. The vSphere api spec is not consistent, and it would be challenging to fix this. Ultimately, the user will still get an error indicating if the parameter is required or not

PR #672, opened 2026-07-31
Regenerate vcenter_ovf_libraryitem for vSphere 9.1.0

mikemorency:
##### SUMMARY

Regenerate the `vcenter_ovf_libraryitem` module using the vSphere 9.1.0 API specification. This modernizes the module to use the shared `ActionModule` base class, adds structured argument specs with proper suboption validation, and includes comprehensive unit and integration tests.

Key changes:
- Regenerated module from the 9.1.0 OpenAPI spec, replacing the legacy monolithic implementation with the `ActionModule` base class
- `deployment_spec.accept_all_EULA` renamed to `deployment_spec.accept_all_eula` (following the 9.1.0 spec's snake_case normalization). This is a **breaking change** for existing playbooks
- Added new `subnet_mappings` and `tag_params` suboptions to `deployment_spec` (new in vSphere 9.1.0)
- Bugfixes in the `ActionModule` base class for action-only modules
- Added a legacy output shim for backwards compatibility with module return values
- Added SonarQube duplicate code exclusion for generated modules
- Full unit test suite and MockServer-based integration tests

> [!NOTE]
> This PR body was generated with the help of AI

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vcenter_ovf_libraryitem

##### ADDITIONAL INFORMATION
The `accept_all_EULA` to `accept_all_eula` rename follows the vSphere 9.1.0 API spec. Existing playbooks using `accept_all_EULA` will need to update to `accept_all_eula`.  This module is *not* backwards compatible.

```paste below
Breaking change example - before:
  deployment_spec:
    accept_all_EULA: true

After:
  deployment_spec:
    accept_all_eula: true
```


plugins/module_utils/_crud_module.py
```diff
@@ -123,48 +150,82 @@ def ensure_present(self) -> dict:
...
+        try:
+            # brittle attempt at maintaining backwards compat idempotency. This really is an edge case, in an edge case,
+            # that would be inconvinient if broken.
+            error_type = result["value"]["error"]["errors"][0]["error"]["error_type"]
+            if error_type == "ALREADY_EXISTS":
+                result["diff"] = {}
+                result["changed"] = False
+                return
+
+        except KeyError:
```
p3ck: You should probably add IndexError to this except as well since I don't think KeyError will catch if errors returns an empty array.

plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -1,636 +1,778 @@
...
-
-async def entry_point(module, session):
-
-    if module.params["state"] == "present":
-        if "_create" in globals():
-            operation = "create"
+    try:
+        if module.params["state"] == "present":
+            result = crud_module.ensure_present()
+        elif module.params["state"] == "absent":
```
p3ck: is this dead code?  I don't see absent as an option above.

plugins/modules/vcenter_ovf_libraryitem.py
```diff
@@ -1,636 +1,778 @@
...
-    get_device_info,
-    get_subdevice_type,
-    open_session,
-    prepare_payload,
-    session_timeout,
-    update_changed_flag,
+MOID_PARAMETER_HINTS = ["ovf_library_item_id"]
+
+LIST_ENDPOINT = "/vcenter/ovf/library-item"
+ITEM_ENDPOINT = ""
```
p3ck, line 427: This is never used as far as I can tell

plugins/modules/vcenter_ovf_libraryitem.py
mikemorency: yea, its from the generation script. I fixed it

plugins/modules/vcenter_ovf_libraryitem.py
mikemorency: its an artifact of the generation. I dont think it will be easy to remove

p3ck (approved):

mariolenz:
It's somewhat weird that there are changes to `vcenter_datacenter` and `vcenter_resourcepool` if you just want to regenerate `vcenter_ovf_libraryitem`. This looks somewhat suspicious to me.

Anyway, can't we replace this module with something else instead of regenerating it? I've just had a quick glance so I'm not really sure. There might not be a possible replacement in [community.vmware](https://github.com/ansible-collections/community.vmware) (at least none that isn't deprecated) but what about [vmware.vmware.import_content_library_ovf](https://github.com/mariolenz/vmware.vmware/blob/main/plugins/modules/import_content_library_ovf.py)? Isn't this a substitute? Or can it be made into one?

Apropos of nothing, reviewing this AI generated stuff is a pain in the ass.

mikemorency:
@mariolenz Since this is only the third module to be regenerated (i group the crud/info module pairs as one), theres still bugs being found. Hence the changes to resourcepool/datacenter; they are bugfixes identified when testing this module. Or rather, those changes are made to ensure backwards compatibility which was identified as lacking while testing this module.

I am hopeful the PRs get simpler and simpler. Theres a ton of module here so im trying to take it slow for now and generate different "types" to get as much of the bugs out as possible in the beginning.

As for replacing this module, i think bits and pieces are covered by other modules but not all of it. Once I am done with thie regeneration, I was going to come back to it anyway since its a breaking change. Ill look and see if its worth migrating/replacing completely then. I cant just not do anything with it, since it wasnt previously deprecated and everything is being generated based on vsphere 9

mariolenz:
@mikemorency Thanks for explaining! And sorry for the "pain in the ass" :pray:  I guess this sounds a little bit harsh. Maybe just "hard" would have been better. And it's really hard for me because I don't find the time to understand generating code with AI, and therefor probably just don't have enough experience.

> As for replacing this module, i think bits and pieces are covered by other modules but not all of it.

No problem there. I just wanted to ping you. After all, it _might_ have been possible to deprecate this module in favor of another one. If there would have been an alternative, it might have been easier for you just to deprecate the module instead of regenerating it. But if there's no viable alternative ATM I think you've done the right thing here.

PR #673, opened 2026-08-04
Regenerate appliance_shutdown modules for vSphere 9.1.0

mikemorency:
##### SUMMARY

Regenerate the `appliance_shutdown` and `appliance_shutdown_info` modules using the new LLM-based generation pipeline targeting the vSphere 9.1.0 API spec. This replaces the legacy async/aiohttp implementation with the synchronous base class pattern (`VmwareRestCrudModuleBase` / `VmwareRestInfoModuleBase`), bringing these modules in line with the rest of the collection.

Also includes supporting fixes to the shared base classes and generation script:
- `_client.py`: Make `data` optional in `Client.post()` for action endpoints with no request body (e.g., cancel).
- `_crud_module.py`: Simplify action routing to pass params directly instead of requiring a resource lookup; handle empty `result[\"value\"]` in success validation.
- `_info_module.py`: Make `list_operation_config` optional for info modules that only support GET (no LIST endpoint).
- `generate_module.py`: Conditionally include `list_operation_config` in info module `main()` only when a LIST operation exists.
 
Unit tests and MockServer-based integration tests are included for both modules.

> *This PR description was generated with the help of AI.*

##### ISSUE TYPE
- Feature Pull Request
 
##### COMPONENT NAME
- `appliance_shutdown`
- `appliance_shutdown_info`
- `plugins/module_utils/_client.py`
- `plugins/module_utils/_crud_module.py`
- `plugins/module_utils/_info_module.py`
- `content_generation/generate_module.py`


Yaish25491:
Summary

  This PR replaces the legacy async/aiohttp appliance_shutdown and appliance_shutdown_info modules with the new synchronous VmwareRestCrudModuleBase / VmwareRestInfoModuleBase pattern, targeting vSphere 9.1.0. It also modifies four
  shared base class files (_client.py, _crud_module.py, _info_module.py, generate_module.py) to support action-only modules (no create/update/delete), info modules without a LIST endpoint, and POST endpoints with no request body.
  Unit tests and MockServer integration tests are included. The module rewrites are clean and well-structured, but two base class changes warrant careful review for cross-module impact.

  Findings

  Blockers (must fix before merge)

  - [BACKWARDS COMPAT] plugins/module_utils/_crud_module.py:116–122 — perform_action() replaces self._resolve_resource_context() with self.params, removing resource lookup for all modules that use action operations, not just
  appliance_shutdown. Any existing module that uses perform_action() with a path-templated action URI (e.g., /vcenter/vm/{vm}/power?action=start) previously relied on _resolve_resource_context() to resolve the resource MOID from a
  name lookup. After this change, modules must receive the MOID directly in params or build_path() raises RequiredPathParameterError. If no other modules in the collection currently use perform_action(), this is safe — but that
  assumption should be validated before merging. The "No matching resource was found" error message is also removed, giving users a less descriptive RequiredPathParameterError on missing MOID.
  - [DOCUMENTATION] plugins/module_utils/_crud_module.py:143–146 — The comment "Validation of MOIDs in the URI… left up to the vSphere API" is added to the ensure_present() docstring, but ensure_present() still calls
  _resolve_resource_context() and resolves MOIDs itself. The comment belongs in perform_action(), which is the method where the resource lookup was actually removed.

  Warnings (should fix, not strictly blocking)

  - [IDEMPOTENCY/UX] plugins/modules/appliance_shutdown.py:168–172 — The delay and reason parameters are documented as "Required when state=poweroff or state=reboot" but the argument spec has no required_if enforcement. Validation
  only happens inside build_body() → _build_dict_with_params(), which raises RequiredParameterError. Using Ansible's built-in required_if=[['state', 'poweroff', ['delay', 'reason']], ['state', 'reboot', ['delay', 'reason']]] would
  give a more user-friendly error message and catch the issue at argument validation time rather than at API request construction time.
  - [BACKWARDS COMPAT] plugins/modules/appliance_shutdown.py:271 / appliance_shutdown_info.py:100 — version_added changed from 2.0.0 to 5.0.0. Since these modules are being regenerated (not newly created), existing users upgrading
  from collection v2.x–v4.x will see modules they've been using appear to be "new" in the docs. Consider keeping version_added: 2.0.0 or adding a note that these are regenerated replacements.
  - [CODE QUALITY] plugins/modules/appliance_shutdown.py:173 — requirements: [] is unusual. The old module required aiohttp and python >= 3.6. The new module has no requirements, which is correct since it uses the built-in urllib,
  but the empty list is noise — either omit the field entirely or add a meaningful entry.

  Suggestions (optional improvements)

  - [TESTING] tests/unit/plugins/modules/test_appliance_shutdown.py — Tests only verify mock_client.post.assert_called_once() and body content. Consider asserting the POST URL path matches the expected action URI (e.g.,
  /appliance/shutdown?action=reboot) to catch regressions if the OperationConfig URIs change.
  - [CODE QUALITY] plugins/module_utils/_info_module.py:41–49 — The return self.normalize_info_results(...) for the list fallback was moved outside the except block. While functionally correct (the try block always returns), the
  indentation change makes the control flow less obvious to readers. A comment like # Fall through to list operation above the return would help.
  - [TESTING] tests/unit/plugins/module_utils/test_crud_module.py:370–377 — The renamed test test_perform_action_missing_path_parameter now tests a different behavior than before (path parameter missing vs. resource not found). The
  old test for "resource not found → fail_json" should be kept alongside the new test if _resolve_resource_context() is used by other action modules, or explicitly removed with a comment explaining the design change.

mikemorency:
@Yaish25491 some of the suggestions are not relevant, but i fixed some of the test cases and updated documentation as suggested

Yaish25491 (approved):
lgtm

PR #686, opened 2026-09-03
feat: regenerate appliance modules for vSphere 9

mikemorency:
##### SUMMARY
Regenerates the remaining vSphere `appliance` REST modules against the vSphere 9 API spec. Alongside the regeneration this cleans up a few long-standing `state` inconsistencies and argument-spec gaps, and adds integration/unit test coverage for the affected modules.

The following modules were regenerated:

- `appliance_networking`
- `appliance_networking_info`
- `appliance_networking_interfaces_info`
- `appliance_networking_interfaces_ipv4`
- `appliance_networking_interfaces_ipv4_info`
- `appliance_networking_interfaces_ipv6`
- `appliance_networking_interfaces_ipv6_info`
- `appliance_shutdown`
- `appliance_system_storage`
- `appliance_system_storage_info`
- `appliance_system_time_info`
- `appliance_system_version_info`
- `appliance_update_info`

> **Note:** The `vmon` service modules are intentionally **not** included here and will need to be regenerated separately, since they do not have vSphere 9 support.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
appliance

##### ADDITIONAL INFORMATION
Notable behavior changes captured in the changelog fragments:

- **breaking_changes**
  - `appliance_networking_interfaces_ipv4` / `appliance_networking_interfaces_ipv6` - `state` now accepts `present` instead of `set`.
  - `appliance_system_storage` - `state` now accepts `resize-ex` instead of `resize_ex`, to align with the API.
- **minor_changes**
  - `appliance_networking_interfaces_info` - the redundant `id` return value (which only echoed the `interface_name` input) is no longer emitted.
- **bugfixes**
  - `appliance_networking` - add `present` to the `state` choices so the `ipv6_enabled` update path is reachable.
  - `appliance_networking_interfaces_ipv4_info` / `appliance_networking_interfaces_ipv6_info` - mark `interface_name` as required so the module fails clearly during argument validation.

```paste below

```


Yaish25491 (approved):
### Summary

Regenerates 13 `appliance` modules onto the new `OperationConfig`/`VmwareRestCrudModuleBase` framework against the vSphere 9.1.0 spec, plus three small generator fixes, 13 new unit test files (734 tests) and 8 new integration targets. The mechanical quality is high: `ansible-test sanity` (validate-modules, pep8, pylint, yamllint) passes clean and all 734 unit tests pass locally. Documentation is a clear improvement over the old autogenerated blocks, the `resize-ex` and `state: set → present` breaking changes are correct against the real spec, and the test targets follow existing repo convention.

The problem is that two modules cannot work against a real vCenter, and the test fixtures hide it. Both `appliance_networking` and `appliance_networking_interfaces_ipv6` diff the desired body against a GET response whose schema does not match, and in both cases the mock GET responses were written to match the desired body rather than the API. I verified this against `content_generation/api_specs/9.1.0/automation/vcenter.json`, which is bundled in this repo.

### Findings

#### Blockers (must fix before merge)

- **[IDEMPOTENCY] `plugins/modules/appliance_networking.py:110`** — `ipv6_enabled` can never be applied; the module always reports `changed: false` and never sends the PATCH. The GET at line 104 returns `Appliance.Networking.Info`, whose schema is `required: [dns, interfaces]` with exactly those two properties — `ipv6_enabled` exists *only* in `Appliance.Networking.UpdateSpec`. `_crud_module.py:89` returns `{**self.params, **api_response}`, so `_calculate_resource_diff` reads `current.get("ipv6_enabled")` back out of the user's own params, compares it to itself, and finds no diff. This is exactly the behaviour `changelogs/fragments/09022026-appliance-networking-fixes.yml` claims to fix. Neither test catches it because both fixtures invent the field: `tests/unit/plugins/modules/test_appliance_networking.py:69` and `tests/integration/targets/appliance_networking/openapi_spec_mocks/default.json:31`. The fix needs to live in the diff logic (e.g. treat a body key absent from the raw API response as "unknown, must write") rather than in the module.

- **[IDEMPOTENCY] `plugins/modules/appliance_networking_interfaces_ipv6.py:162`** — `addresses` never compares equal, so every run reports `changed: true` and issues a PUT. The GET returns `Ipv6.AddressInfo` (`required: [address, origin, prefix, status]`); the PUT body is built from `Ipv6.Address` (`address`, `prefix` only). `_crud_module.py:244` `_values_equal` only recurses into dicts — a list falls through to `==`, so `[{address, prefix}] != [{address, prefix, origin, status}]` on every invocation. The module's own sibling docs confirm the real shape (`appliance_networking_interfaces_ipv6_info.py:60` samples `origin: STATIC, status: PREFERRED`). Masked by `tests/unit/plugins/modules/test_appliance_networking_interfaces_ipv6.py:79` and `tests/integration/targets/appliance_networking_interfaces_ipv6/openapi_spec_mocks/default.json:42`, neither of which includes `origin`/`status`. Violates architectural invariant #1.

#### Warnings (should fix, not strictly blocking)

- **[CHANGELOG] `changelogs/fragments/09022026-appliance-networking-fixes.yml:3`** — factually wrong. `main`'s `appliance_networking` already ships `"choices": ["present", "reset"], "default": "present"`; `state` did not "only accept `reset`". The generator was missing `present` for update-only endpoints, but that never reached the released module. The trailing "leaving the IPv4/IPv6 update path unreachable" is also wrong — this module only toggles `ipv6_enabled`.

- **[DOCUMENTATION] `appliance_networking_interfaces_ipv4_info.py:55` and `appliance_networking_interfaces_ipv6_info.py:55`** — RETURN documents an `id` key the module never emits (their own unit tests assert `"id" not in result`, because the GET response has no `interface_name` attribute for `_get_moid_attribute_value_from_resource` to find), and omits `info`, which `normalize_info_results` *always* returns. Every other info module in this PR documents `value` + `info`.

- **[DOCUMENTATION] `appliance_shutdown.py:87`** — the RETURN block drops the `id` key, but `perform_action()` (`_crud_module.py:120`) still sets it. Undocumented return value, and no changelog entry for the doc removal.

- **[ERROR HANDLING] argspec vs. body_spec requiredness** — `appliance_networking_interfaces_ipv4.py:53` documents `mode` as `required: false` while `body_spec` (line 150) and `Ipv4.Config` require it; `appliance_networking_interfaces_ipv6.py` does the same for all four of `dhcp`/`autoconf`/`addresses`/`default_gateway`; `appliance_shutdown.py` states `delay`/`reason` requirements only in prose. Users get a `RequiredParameterError` after a live GET round trip instead of argspec-time validation, and the published docs say the parameters are optional. `required_if` — already used by ~10 modules in this collection — covers the shutdown and ipv4/ipv6 cases cleanly.

- **[DOCUMENTATION] `appliance_networking_interfaces_ipv4.py:46`** — `mode` has no `choices`, though the description enumerates DHCP/STATIC/UNCONFIGURED and the 9.1.0 spec defines exactly those three. A typo currently reaches the API.

- **[BACKWARDS COMPATIBILITY] `changelogs/fragments/09022026-appliance-networking-interfaces.yml:8`** — removing the `id` return value from `appliance_networking_interfaces_info` is filed under `minor_changes`. Removing a return value breaks any playbook referencing `result.id`; this belongs in `breaking_changes` alongside the others.

#### Suggestions (optional improvements)

- **[TESTING] `openapi_spec_mocks/*.json`** — deriving the mock GET examples from the response schemas rather than the request bodies would have caught both blockers on its own. Concretely: add `origin`/`status` to the ipv6 address entries, and drop the invented `ipv6_enabled` from the networking GET example.
- **[TESTING] `content_generation/`** — no unit tests exist for the generator, so all three changes here (`merge_parameters` exclusion, `has_item_update`, the `no_log` heuristic) ship untested. The `no_log` heuristic at `generate_module.py:789` also matches non-secrets like `password_expires_at`, which would silently hide a legitimate value; emitting `no_log: false` for a known allow-list would keep validate-modules happy without over-redacting. Note this heuristic is currently exercised by nothing in this PR — none of the 13 regenerated modules has a `password`/`secret` parameter.

### Checklist Status

| Category                | Status | Notes                                                                                                                                                                    |
| ----------------------- | ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Collection Metadata     | PASS   | `galaxy.yml` 5.0.0 matches `version_added: 5.0.0`; all modules already in the `vmware_rest` action group; no new deps.                                                   |
| Module Documentation    | FAIL   | Undocumented `info`/`id` mismatches on ipv4_info/ipv6_info and shutdown; `mode` missing `choices`; required-ness understated.                                            |
| Naming and Style        | PASS   | Filenames and module names unchanged; consistent with merged new-style modules.                                                                                          |
| Idempotency             | FAIL   | Two modules non-functional/non-idempotent against the real API (blockers 1 and 2).                                                                                       |
| Sensitive Data          | PASS   | Only `vcenter_password` from the shared fragment; `no_log=True` set there. No secrets in returns.                                                                        |
| Error Handling          | PASS   | All failures route through `module.fail_json()` via `VmwareModuleError`; no bare `raise`/`sys.exit()`; shared `module_utils` used throughout.                            |
| Testing                 | FAIL   | 734 unit tests and 8 integration targets pass, but two idempotency assertions are validated against fixtures that contradict the API schema; generator changes untested. |
| Backwards Compatibility | PASS   | `state: set → present` and `resize_ex → resize-ex` verified correct against the 9.1.0 spec and documented as breaking; the `id` removal is miscategorised (see warning). |
| Changelog Fragment      | FAIL   | The `appliance_networking` bugfix entry describes a bug that does not exist in the shipped module.                                                                       |
| Code Quality            | PASS   | No dead code or debug statements; sanity (pep8, pylint, yamllint, validate-modules) clean; no injection or credential issues.                                            |

### Verdict

**REQUEST CHANGES** — `appliance_networking` cannot apply `ipv6_enabled` at all and `appliance_networking_interfaces_ipv6` writes on every run, and in both cases the test fixtures were written to match the request body rather than the documented response schema, so CI will stay green while the modules are broken.

PR #684, opened 2026-09-01
Regen appliance_monitoring_info and appliance_monitoring_query modules

mikemorency:
##### SUMMARY
Regenerate modules `appliance_monitoring_info` and `appliance_monitoring_query` for retrieving vSphere appliance monitoring data and querying monitoring metrics via the vCenter REST API. Modules were generated from the OpenAPI spec, with accompanying integration and unit tests.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
appliance_monitoring_info, appliance_monitoring_query

##### ADDITIONAL INFORMATION
These are unusual API endpoints so some bugfixes were made as well

Assisted-by: Claude Sonnet 5 (claude-sonnet-5)

PR #683, opened 2026-08-31
Regenerate appliance local accounts modules for vSphere 9.1.0

mikemorency:
##### SUMMARY
Regenerates the appliance local accounts modules for vSphere 9.1.0 and adds test coverage.

- Regenerate `appliance_localaccounts_info`, `appliance_localaccounts_globalpolicy`, and `appliance_localaccounts_globalpolicy_info` using the current generation tooling.
- Minor supporting updates to `content_generation/generate_module.py`, `plugins/module_utils/_client.py`, and `plugins/module_utils/_info_module.py`.
- Add unit tests for all three modules.
- Add integration test targets (MockServer-based) for `appliance_localaccounts` and `appliance_localaccounts_globalpolicy`, covering list/get, idempotency, and check mode.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
appliance_localaccounts_info
appliance_localaccounts_globalpolicy
appliance_localaccounts_globalpolicy_info

##### ADDITIONAL INFORMATION
Integration tests pass against the MockServer simulator:

```paste below
make integration INTEGRATION_TARGETS=appliance_localaccounts
...
PLAY RECAP
testhost : ok=24 changed=2 unreachable=0 failed=0 skipped=11 rescued=0 ignored=0
```

PR #687, opened 2026-09-03
Regenerate content modules for vSphere

mikemorency:
##### SUMMARY
Regenerated the set of `content` modules from the vSphere OpenAPI specification, refreshing their generated documentation, argument specs, and payload handling. This PR also adds the accompanying generated unit and integration test suites (backed by the OpenAPI MockServer simulator) for the regenerated modules.

The following modules were regenerated:

- `content_configuration`
- `content_configuration_info`
- `content_library_info`
- `content_library_subscriptions_info`
- `content_locallibrary_info`
- `content_subscribedlibrary_info`

No changelog fragment is included since these modules are completely backwards compatible.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
content_configuration
content_configuration_info
content_library_info
content_library_subscriptions_info
content_locallibrary_info
content_subscribedlibrary_info

##### ADDITIONAL INFORMATION
Modules were regenerated using the collection's content generation workflow. New unit tests (`tests/unit/plugins/modules/`) and integration tests (`tests/integration/targets/`) were added for the regenerated modules.


Yaish25491 (approved):
### Summary

Regeneration of six `content_*` modules onto the collection's new `OperationConfig` / `VmwareRestCrudModuleBase` architecture, plus 2,140 lines of new unit tests and five MockServer-backed integration targets. The work is high quality and follows the precedent set by the already-merged regen PRs (#683, #684): `version_added: 5.0.0`, `Ansible Eco Content Team` author, `notes: Generated from vSphere API spec 9.1.0`, aliases/meta layout, and `prepare_simulator` usage all match. I verified locally against the PR head: **102 unit tests pass** and **`ansible-test sanity` passes clean** on all six modules (ansible-core 2.17, Python 3.11); all 40 CI checks on the PR are green. The input-compatibility claim in the PR description holds — I diffed every documented option old-vs-new and no parameter was removed, renamed, or had its type or `required` flag tightened.

The problems are concentrated in the RETURN documentation, which was not validated against what the modules actually emit, and in the fidelity of two of the five mock specs.

### Findings

#### Blockers (must fix before merge)

- **[MODULE DOCUMENTATION] `plugins/modules/content_library_info.py:66`, `content_locallibrary_info.py:67`, `content_subscribedlibrary_info.py:67` — the documented `id` return key is never emitted.** All three declare `id` as `returned: When only one library, identified by O(library_id), was queried`, but that never happens. `normalize_info_results` derives `id` via `_get_moid_attribute_value_from_resource`, which searches `MOID_PARAMETER_HINTS` (`["library_id"]` plus the implicit `"resource_id"`) — while the vSphere `LibraryModel` response keys the identifier as `id`. I confirmed this by running the PR's own `test_get_library_by_id` with an added assertion:

  ```
  RESULT KEYS: ['info', 'value']
  E  AssertionError: RETURN documents id but it was not emitted
  E  assert 'id' in {'info': [{'id': 'lib-1001', 'name': 'my_local_library', ...}], 'value': {...}}
  ```

  The payload contains `id: lib-1001` and the module still drops it. Fix either by adding `"id"` to `MOID_PARAMETER_HINTS` (safe — the hints list is only consumed by `_get_moid_attribute_value_from_resource` and `ensure_present`'s exception branch, not by path building) or by deleting the `id` key from the three RETURN blocks. Note that `content_library_subscriptions_info` is *not* affected: its hints include `subscription`, which the response does carry, and `test_content_library_subscriptions_info.py:89` asserts it.

  This slipped through because no test covers the positive case. The three affected test files only assert `"id" not in result` on the 404 path (`test_content_library_info.py:116`, `test_content_locallibrary_info.py:120`, `test_content_subscribedlibrary_info.py:122`) — which passes for the wrong reason. Whichever fix you pick, please add an assertion on the by-ID path.

#### Warnings (should fix, not strictly blocking)

- **[MODULE DOCUMENTATION] `plugins/modules/content_configuration_info.py:53` — RETURN omits `info`.** `normalize_info_results` always returns `info` alongside `value`, and the target's own integration test asserts on it (`tests/integration/targets/content_configuration/tasks/info.yml`: `content_config.info | length == 1`). The other four info modules in this PR document `info`, as does the merged `appliance_monitoring_query`.

- **[MODULE DOCUMENTATION] `plugins/modules/content_configuration.py:428` — RETURN omits `id` and `diff`.** `ensure_present()` seeds `result` with `{"changed": False, "id": ""}` and sets `result["diff"]`, so both are always returned. `diff` in particular is asserted on by the integration test (`update.yml:44-47`). The merged regen CRUD module `vcenter_datacenter.py` documents `id`; the old `content_configuration` did too, so this is also a small doc-level regression.

- **[ERROR HANDLING] `plugins/modules/content_configuration.py:877` → `plugins/module_utils/_crud_module.py:200` — unhandled `AttributeError` when the GET returns 404 or an empty body.** `ensure_present()` routes to `_create()` when `_resolve_resource_context()` is falsy, but this singleton module defines no `CREATE_OPERATION`, so `create_operation_config` is `None`. Reproduced:

  ```
  >   path = self.create_operation_config.build_path(params=self.params)
  E   AttributeError: 'NoneType' object has no attribute 'build_path'
  ```

  The user gets a traceback rather than `module.fail_json()`, violating architectural invariant #5. This is a pre-existing gap in the shared base class — merged `appliance_localaccounts_globalpolicy.py` has the identical GET+UPDATE-without-CREATE shape — so the fix belongs in `_crud_module.py:ensure_present` (fail with an actionable message when `create_operation_config is None`), not in this module. Flagging because this PR adds a second module with the same exposure.

- **[CHANGELOG] No fragment for `content_configuration`'s 16 new parameters.** The "no fragment needed" rationale is sound for the five info modules and matches #683, which shipped fragment-free. But `content_configuration` gains `automatic_sync_refresh_interval`, `automatic_sync_setting_refresh_interval`, `transfer_throttling_bandwidth_total`, `transfer_nfc_max_concurrent_transfers_per_host`, `priority_transfer_threads_pool_size`, `transfer_threads_pool_size` and ten `*_setting` dicts — new user-facing capability that warrants `minor_changes`. All six modules also flip to `supports_check_mode=True`, which reverses the v4.10.0 entry "Disable check mode for all non-info modules in this collection"; worth a line so users aren't surprised.

- **[TESTING] `tests/integration/targets/content_library/openapi_spec_mocks/*.json` and `content_library_subscriptions/openapi_spec_mocks/*.json` use unedited generator placeholders.** The GET examples are literally `{"name": "my_library"}` and `{"name": "my_subscriptions"}`, so `content_library/tasks/info.yml:47` and `content_library_subscriptions/tasks/info.yml:52` assert on a synthetic `name` field that the real `SubscriptionInfo` response does not even have. Contrast the sibling targets: `content_locallibrary` and `content_subscribedlibrary` mocks were hand-curated to realistic `LibraryModel` payloads (`id`, `type`, `storage_backings`, `subscription_info`) that match the modules' RETURN samples. Since `/content/library/{library_id}` returns the same `LibraryModel` as `/content/local-library/{library_id}`, the difference reads as an oversight. Curating these two would also have caught the `id` blocker above.

#### Suggestions (optional improvements)

- **[TESTING] `content_library/openapi_spec_mocks/list_multiple.json`** returns list entries as dicts keyed `library_id`, which drives `_list_resource_details` down the summary-enrichment branch. Real vSphere returns bare identifier strings, exercising the `else` branch. The unit test `test_list_all_libraries` covers the string case correctly, so this is only an integration-side gap.
- **[TESTING] `content_library/tasks/info.yml:82-89`** asserts only `| length == 2`. Because both mock items resolve to the same detail payload, a per-item mix-up would pass. Give the two `library_id` values distinct GET responses and assert on names.
- **[IDEMPOTENCY] `content_configuration/tasks/update.yml:49-80`** labels test 2 "idempotent" but swaps the mock spec to `updated.json` rather than re-running the same task against persisted state. Correct given MockServer's statelessness, and the unit test `test_update_configuration_idempotent` covers the real logic — worth a comment noting the limitation.
- **[CODE QUALITY] `plugins/modules/content_configuration.py:52-404`** exposes ten `*_setting` dicts that the module's own docs describe as "Read-only metadata". They are genuinely optional-on-update in the 9.1.0 `Content.ConfigurationModel` schema, so generating them is faithful — but each requires all three suboptions (`name`, `reboot_required`, `constraints`) when supplied, and they feed the diff calculation. Given the precedent of removing the non-operational `label` parameter (fragment `08112026-vcenter_vm_hardware_floppy.yml`), consider whether these belong in the argspec at all.
- **[CODE QUALITY] `plugins/module_utils/_info_module.py:95`** merges the enrichment key into each list-mode result, so `info[i]` carries a synthetic `library_id` in list mode that `value` does not carry in get-mode (confirmed by `test_content_library_info.py:158`). Undocumented shape inconsistency; out of scope for this PR but relevant to the RETURN samples.

### Checklist Status

| Category                | Status | Notes                                                                                                                                           |
| ----------------------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------------------- |
| Collection Metadata     | PASS   | `meta/runtime.yml` action_group already lists all six; no routing/deprecation conflicts; no new deps                                            |
| Module Documentation    | FAIL   | `id` documented but never returned (×3); `info` and `id`/`diff` returned but undocumented                                                       |
| Naming and Style        | PASS   | Matches merged regen modules exactly                                                                                                            |
| Idempotency             | PASS   | `_calculate_resource_diff` verified by unit + integration tests                                                                                 |
| Sensitive Data          | PASS   | Only `vcenter_password` is sensitive; `no_log=True` set in the shared spec; no new secrets                                                      |
| Error Handling          | FAIL   | Unhandled `AttributeError` on the 404/empty-GET path in `content_configuration`                                                                 |
| Testing                 | PASS   | 102 unit tests + `ansible-test sanity` pass locally; all 40 CI checks green; mock fidelity gaps noted above                                     |
| Backwards Compatibility | PASS   | No parameter removed, renamed, or tightened; `value` shape preserved; connection params moved to the doc fragment, not dropped                  |
| Changelog Fragment      | FAIL   | Absent; `content_configuration`'s 16 new parameters need `minor_changes`                                                                        |
| Code Quality            | PASS   | Generator refactor is clean — `_create_path_parameters` correctly derives nested path params, and all nine call sites were updated consistently |

### Verdict

**REQUEST CHANGES**

Solid, well-tested regeneration that is genuinely input-compatible as claimed, but three modules ship a RETURN key they never emit, and the new-parameter surface on `content_configuration` needs a changelog fragment.

PR #695, opened 2026-09-21
regenerate the remaining modules

mikemorency:
##### SUMMARY
Regenerate the remaining modules

vcenter_vm_hardware
vcenter_vm_hardware_info
vcenter_vm_hardware_adapter_sata_info
vcenter_vm_hardware_adapter_scsi_info
vcenter_vm_hardware_boot_device
vcenter_vm_hardware_boot_device_info
vcenter_vm_hardware_boot_info
vcenter_vm_hardware_boot
vcenter_vm_hardware_cdrom_info
vcenter_vm_hardware_disk_info
vcenter_vm_hardware_ethernet_info
vcenter_vm_hardware_parallel_info
vcenter_vm_hardware_parallel
vcenter_vm_hardware_serial_info
vcenter_vm_hardware_serial
vcenter_vm_tools_installer_info
vcenter_vm_tools_installer

##### ISSUE TYPE
- Module Regen Pull Request

I did not get to generate the integration tests for these. Ill do that later


plugins/module_utils/_crud_module.py
```diff
@@ -230,6 +235,17 @@ def _create(self) -> tuple:
 
         return new_id, value
 
+    def _update_without_idempotence(self):
+        path = self.update_operation_config.build_path(params=self.params)
```
p3ck, line 243: Unlike `_update_if_needed` (line 250), `_update_without_idempotence` has no `if self.update_operation_config is None` guard. If a list-returning GET endpoint were ever configured without an `update_operation_config`, this dereferences `None.build_path(...)` and raises `AttributeError`. All current list modules happen to carry an update config, but the asymmetry is a latent crash; add the same guard for consistency.

plugins/modules/vcenter_vm_hardware.py
```diff
@@ -1,444 +1,273 @@
...
-            fallback=(env_fallback, ["VMWARE_REST_LOG_FILE"]),
-        ),
-        "session_timeout": dict(
-            type="float",
-            required=False,
-            fallback=(env_fallback, ["VMWARE_SESSION_TIMEOUT"]),
-        ),
+def create_module_argument_spec() -> dict:
+    module_args = connection_params_argument_spec()
+    module_args["upgrade_policy"] = {
```
p3ck, line 222: The regenerated argument spec drops the `choices` lists the previous module enforced for `upgrade_policy` (AFTER_CLEAN_SHUTDOWN/ALWAYS/NEVER), `upgrade_version`, and `version` (VMX_*). Invalid values are no longer rejected up front by Ansible and are passed straight to vCenter, which fails with a less clear error. DOCUMENTATION still enumerates the valid values, so this is a narrowed validation vs. prior behavior.

plugins/module_utils/_crud_module.py
```diff
@@ -178,6 +178,11 @@ def ensure_present(self) -> dict:
             result["id"] = new_id
             result["value"] = value
             result["changed"] = True
+        elif not isinstance(resource, dict):
```
p3ck, line 182: This new branch (and the `if not resource:` create branch above) mishandles singleton endpoints whose GET returns a bare JSON list, e.g. `vcenter_vm_hardware_boot_device` whose boot order can legitimately be empty. When the current order is `[]`, control never reaches this `elif`: the empty list is falsy, so `if not resource:` above treats it as "resource not found" and — because these modules have no `create_operation_config` — `fail_json`s with "could not be found ... does not support creating it" instead of doing the PUT update. That breaks the module's primary use case of setting the boot order on a VM that currently has none. Distinguish a real 404/None from an empty-but-present list.

plugins/module_utils/_crud_module.py
```diff
@@ -178,6 +178,11 @@ def ensure_present(self) -> dict:
...
+            result["id"] = self._get_moid_attribute_value_from_resource(self.params)
```
p3ck, line 183: Note: this new branch relies on `_resolve_live_resource_context()` returning a `list`, but that method is still annotated `-> Union[dict, None]` with a docstring saying it returns "an empty dict ... otherwise the resource state." Worth updating the annotation/docstring so the list-shaped return is documented for future callers.

plugins/modules/vcenter_vm_hardware.py
mikemorency: I do not think this one is worth attempting to fix as this is the last batch of modules and it would required messing with the generation script

plugins/module_utils/_crud_module.py
mikemorency: switched most empty responses to None, it seems like that works with all the current modules

p3ck (approved):

